Issue #4, opened 2015-11-12
Dockerfile conflict + outdated(?) shell script


dsifford:
Hi there. 

First of all, excellent base image! I read your associated blog post and the way you have this set up is extremely clever.

Some issues I'm having are the following...

1. The `install-plugins` shell script you have in this repository is outdated and does not work. I think [this line](https://github.com/visiblevc/wordpress-starter/blob/master/scripts/install-plugins.sh#L9) needs to be updated to this....

```sh
curl -LOk# https://downloads.wordpress.org/plugin/${plugin}.zip
```

2. Is the Dockerfile in this repo the same as the one used in your dockerhub image? The reason I ask is because I can cleanly install wordpress using your dockerhub image, but I can not do the same if I fork this repo, and change the `image: visiblecv/wordpress` line in the docker-compose.yml file to `build: .`. Could this an error on my end?

Thanks for the help!

karellm:
@dsifford Thanks for reporting this, I will investigate over the weekend.

karellm:
@dsifford I will close this for now. I pushed the plugin fix.

PR #8, opened 2015-11-14
[Fix] - DOS/UNIX Line Endings Error

dsifford:
- For some users (tested on Windows) the Dockerfile would not run to completion
due to DOS line endings in the .sh files confusing UNIX VMs. This commit adds the
dos2unix package and converts the files on the fly during the build of the
Dockerfile.
- Secondarily, this PR also adds `TERM=xterm` to the build environment to prevent the myriad error messages (see below) that occur without it being explicitly set (at least on windows machines running Docker Toolbox).
- Fixes visiblevc/wordpress-starter#4

```sh
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (This frontend requires a controlling tty.)
debconf: falling back to frontend: Teletype
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@ FROM tutum/apache-php
 
 # Install mysql-client
 RUN apt-get update && apt-get -y upgrade
-RUN apt-get install -y mysql-client unzip
+RUN apt-get install -y mysql-client unzip dos2unix
+ENV TERM xterm
 
 # Install wp-cli
 ADD https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar /usr/local/bin/wp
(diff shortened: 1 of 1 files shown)


karellm:
@dsifford Thanks for the PR, I will add the `ENV TERM` to silence the error messages. But I'm not comfortable adding OS-specific packages and logic to the image. It defeats the purpose of Docker and if we start doing this, it will grow out of control. I'd much rather fix the problem at the core.

I'm not on Windows, nor is anyone that worked on this project. I'm surprised there would be DOS line endings. If you are on Windows, it is more likely that you forked the project, opened it in an editor that changed the line endings. I added an `.editorconfig` file to avoid that (make sure you have the plugin for your editor installed). Can you confirm me that you still have the issue when you make sure that your eol is set to `lf`?

dsifford:
No worries! I got it fixed up for myself so I figured I'd toss it your way for giggles.

I use Atom editor and I set all the shell script files explicitly to `lf` and I still had the issue. Not sure why (although I've got a hunch that it's because I have `git config --global core.autocrlf` set to true). 

A less intrusive method of fixing this (and, honestly, probably a much faster way) is to use `tr '\n' ' ' < input_filename` or `sed ':a;N;$!ba;s/\n/ /g' file` ([found here](http://stackoverflow.com/questions/1251999/how-can-i-replace-a-newline-n-using-sed)) -- I'll likely end up making this switch to with version to keep the Dockerfile as light as possible.

Thanks again brother!

karellm:
@dsifford I close this for now, feel free to open a new one if you find the source of the problem.

PR #9, opened 2015-11-19
Let the containers only listen on localhost

saschalalala:
This way the containers can be run "as is" for development on the localhost without being accessible from the public.


karellm:
This will not work with boot2docker or docker-machine. You do not connect to `127.0.0.1` but your machine IP. I added it for the database though. I invite you to configure the `docker-compose.yml` for the rest.

PR #15, opened 2016-03-05
Remove depreciations, add new compose format,  & shrink Dockerfile

dsifford:
Hi there,

This PR updates this repo for Docker Compose 1.6.x. Specifically...
- Remove depreciated network environment variables (eg `$MYSQL_ENV_MYSQL_ROOT_PASSWORD`, etc).
- Use new compose file format.
- Shrink your Dockerfile to remove redundant layers.
- Add files from this repo into the Dockerfile using curl so that you can set up an automated build on Dockerhub if that's something that interests you.

Cheers :beers:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,27 @@
 FROM tutum/apache-php
+ENV DEBIAN_FRONTEND noninteractive
 
 # Install mysql-client
-RUN apt-get update && apt-get -y upgrade
-RUN apt-get install -y mysql-client unzip
-ENV TERM xterm
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        mysql-client \
+        unzip \
+    && rm -rf /var/lib/apt/lists/*
 
-# Install wp-cli
-ADD https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar /usr/local/bin/wp
-RUN chmod +x /usr/local/bin/wp
-
-# Configure Apache
-RUN sed -i "s/AllowOverride None/AllowOverride All/g" /etc/apache2/apache2.conf
-RUN a2enmod rewrite
-
-# Add some scripts for development
-RUN mkdir -p /scripts
-ADD ./scripts/setup-database.sh /scripts/setup-database.sh
-ADD ./scripts/setup-wordpress.sh /scripts/setup-wordpress.sh
-ADD ./scripts/backup-db.sh /scripts/backup-db.sh
-ADD ./scripts/restore-db.sh /scripts/restore-db.sh
-ADD ./scripts/install-plugins.sh /scripts/install-plugins.sh
-ADD ./scripts/remove-plugins.sh /scripts/remove-plugins.sh
-RUN chmod 755 /scripts/*.sh
+# Install wp-cli, configure Apache, & add scripts
+WORKDIR /app
+RUN mkdir /scripts && curl \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,31 +1,27 @@
 FROM tutum/apache-php
+ENV DEBIAN_FRONTEND noninteractive
 
 # Install mysql-client
-RUN apt-get update && apt-get -y upgrade
-RUN apt-get install -y mysql-client unzip
-ENV TERM xterm
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
karellm: I would prefer that on multiple RUN for readability. Is there a reason you made it a single command? Also any reason you removed the `upgrade` part?

dsifford:
@karellm Each run creates a new layer in your build image, so the combination of multiple `RUNS` into a single run command results in a slimmer (and faster-to-build) image altogether.

It's [generally recommended](http://crosbymichael.com/dockerfile-best-practices-take-2.html) to not include the `upgrade` line due to build failure risks and image consistency concerns.

docker-compose.yml
```diff
@@ -1,28 +1,31 @@
...
-  image: mysql:5.7
-  ports:
-    - localhost:3306:3306
-  volumes_from:
-    - data
-  environment:
-    MYSQL_ROOT_PASSWORD: root
+version: '2'
+services:
+  data:
```
karellm: Shouldn't that become a volume?

```
services:
  wordpress:
    volumes:
      - "data:/var/lib/postgres"
volumes:
  data: {}
```

dsifford: It can, but expressing volumes using the top-level volumes key is only required if you plan on using an alternate volume driver

karellm: Ok but aren't "data-only" containers considered a bad practice now? How would you go about writing this without the data container?

karellm:
Thanks for the great contribution, you actually beat me to it :) We updated our dev stack and the project was the last on the list. 

I like the other changes, I agree that trimming the Docker image is a good idea. If you have other Docker recommendations I'd love to hear them!

Before merging, I will want to see what we do with the Volume thing, can you also squash your commits into a single one?

dsifford:
@karellm Sure thing! I have a repo that I'm using that began as an initial fork of this project that implements a few other changes (that you are more than welcome to use). The repo can be found [here](https://github.com/dsifford/wordpress).

Over the last day or so I added some additional changes to my repo that I think would also serve this repo nicely. If you'd like, look it over and let me know if you'd like me to add any of those changes to this PR and I'll do that as well.

Thanks again for this awesome project! :beers:

karellm:
@dsifford I will have a look, I prefer to keep PRs small though. I actually need to merge that so I can run the update ;) As soon as you have squashed the commits I will merge and we can keep the discussion alive.

dsifford:
@karellm Sounds good

dsifford:
@karellm Good to go :+1:

karellm:
Thanks!

karellm:
@dsifford Would you mind opening an issue to discuss the other improvements? I would love to have you collaborate on this if you feel like merging the effort!

dsifford:
@karellm Done.

Of note: Be sure to test this PR yourself to make sure it's working properly. I tested my fork and then copied the relevant changes over to this PR and never got around to actually doing a full test. There shouldn't be any problems, but who knows, I might have made a typo somewhere.

Issue #11, opened 2016-01-16
Access denied


stefanvermaas:
Hi guys,

I came across your wordpress-starter guide and I wanted to try it out. Though I found that simple running `docker-compose up` after cloning the repo doesn't seem to "work".

It fails when the database is initialized.

This is the code output, any ideas how to fix this? Because as you can see the project doesn't use the correct address, but I've got no idea where to change it to the correct IP address.

```code
wordpress_1 | => Setup Database:
wordpress_1 | =====================================
wordpress_1 |   Database Host Address:  172.17.0.3
wordpress_1 |   Database Port number:   3306
wordpress_1 |   Database Name:          wordpress
wordpress_1 |   Database Username:      root
wordpress_1 | =====================================
mysql_1     | 2016-01-16T14:00:22.062675Z 2 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:00:27.068629Z 3 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:00:32.074264Z 4 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:00:37.079987Z 5 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:00:42.085466Z 6 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:00:47.092656Z 7 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:00:52.097648Z 8 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:00:57.102063Z 9 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:01:02.107606Z 10 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
mysql_1     | 2016-01-16T14:01:07.111878Z 11 [Note] Access denied for user 'root'@'172.17.0.2' (using password: YES)
wordpress_1 | Cannot connect to Mysql
```

Cheers!

karellm:
[...]

PR #17, opened 2016-03-11
run.sh rewrite

### LINKED ISSUE #13: wp-content/uploads permission denied when uploading media
O+G are root:root not www-data:www-data How are you guy's handling this?

dsifford:
#### Major Changes
- Direct mysql use replaced entirely by wp-cli
- Two new optional environment variables added: `PLUGINS` and `SEARCH_REPLACE`
    - `SEARCH_REPLACE`: Accepts a comma-separated string in the form of `http://www.currenturl.com,http://newurl:8080,` and automatically replaces the links in the database if one exists in the `/data` directory
    - `PLUGINS`: Accepts a comma-separated list of plugins that you depend on in the form of `plugin-one,plugin-two,plugin-three,http://www.urltoplugin.com/plugin.zip`
- Adjust wp-cli so that it does not use any root-level privileges
- Add a `wp-cli.yml` file on the fly to negate the need for any long wp-cli commands. (Should also fix the issue described in #11 )
- Remove the `sleep 10` commands and instead use a while loop that pings the MySQL container until it's ready to go.
- Create a `.htaccess` file on the fly with wp-cli
- Fix filesystem permissions to allow media uploads. Closes #13 

#### ~~Todo~~
~~Still need to add better error handling for the wp-cli commands. At present, the build continues without warning if an error occurs. (WIP on my fork).~~

**Edit:** (03/11/2016) Error handling improved...

Now, when any critical build errors occur, the Dockerfile exits immediately with an error code, a helpful message, and the line number where the error occurred.


scripts/run.sh
```diff
@@ -1,16 +1,160 @@
...
+	  dbpass: $DB_PASS
+	  dbname: $DB_NAME
+	  dbhost: db:3306
+
+	core install:
+	  url: $([ "$AFTER_URL" ] && echo "$AFTER_URL" || echo localhost:8080)
+	  title: $DB_NAME
+	  admin_user: root
+	  admin_password: $DB_PASS
+	  admin_email: admin@${DB_NAME}.com
```
karellm: Any reason this is hard-coded?

Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/*
 
 # Install wp-cli, configure Apache, & add scripts
+# Lines 15-20 are not needed for build to succeed
```
karellm: I think you can remove this comment and the lines :+1:

scripts/run.sh
```diff
@@ -1,16 +1,160 @@
 #!/bin/bash
```
karellm: As a styleguide, can you follow a 2 spaces indentation? I noticed couple places use single space. Thanks :)

scripts/run.sh
```diff
@@ -1,16 +1,160 @@
...
+				--no-quiet --skip-columns=guid | grep replacement) || \
+				ERROR $((LINENO-2)) "Could not execute SEARCH_REPLACE on database"
+			echo -ne "$REPLACEMENTS\n"
+
+		fi
+
+	else
+		printf "%s\n" \
+			"=> No database backup found. Initializing new database..." " " \
+			"	  DEFAULT DATABASE CREDENTIALS" \
```
karellm: Any reason why we only show credentials on a new DB? Since we configure the `wp-config.php` it feels we should either always show it or never.

scripts/run.sh
```diff
@@ -1,16 +1,160 @@
...
+if [ "$PLUGINS" ]; then
+	printf "=> Checking plugins...\n"
+	while IFS=',' read -ra plugin; do
+		for i in "${!plugin[@]}"; do
+			sudo -u www-data wp plugin is-installed "${plugin[$i]}"
+			if [ $? -eq 0 ]; then
+				printf "=> ($((i+1))/${#plugin[@]}) Plugin '%s' found. SKIPPING...\n" "${plugin[$i]}"
+			else
+				printf "=> ($((i+1))/${#plugin[@]}) Plugin '%s' not found. Installing...\n" "${plugin[$i]}"
+				sudo -u www-data wp plugin install "${plugin[$i]}"
```
karellm: We might want Error management here

scripts/run.sh
```diff
@@ -1,16 +1,160 @@
...
+printf "Done!\n"
+
+# Install Plugins
+if [ "$PLUGINS" ]; then
+	printf "=> Checking plugins...\n"
+	while IFS=',' read -ra plugin; do
+		for i in "${!plugin[@]}"; do
+			sudo -u www-data wp plugin is-installed "${plugin[$i]}"
+			if [ $? -eq 0 ]; then
+				printf "=> ($((i+1))/${#plugin[@]}) Plugin '%s' found. SKIPPING...\n" "${plugin[$i]}"
```
karellm: How about making sure we have the latest version?

scripts/run.sh
```diff
@@ -1,16 +1,160 @@
...
+				sudo -u www-data wp plugin install "${plugin[$i]}"
+			fi
+		done
+	done <<< "$PLUGINS"
 else
-  echo "Database already setup!"
+	printf "=> No plugin dependencies listed. SKIPPING...\n"
+fi
+
+# Temp workaround until --ignore-plugins & --ignore-themes is fixed on wp-cli
```
karellm: I'm fine we do remove them but how about having a `$THEMES` configuration then? I can see why someone would want a default theme.

karellm:
@dsifford Thanks a lot for that, it is a massive step up. I started to update our article series with proper credit ;) I added a bunch of comments. I will also start a PR for the README.

scripts/run.sh
dsifford: No reason other than just getting rid of an additional external file dependency. 

The file won't be able to be volumed in by users due to it being needed early in the build process. I made the defaults fairly sensible though so it shouldn't cause much of an issue. Plus, you can override them at any time by using different options in-line in your wp-cli commands directly.

scripts/run.sh
dsifford: Sure... I had some issues with 2 spaces due to having the indented `cat<-EOF` command. I'll unindent that and switch to spaces. :+1:

scripts/run.sh
dsifford: Fair point. Though the only thing that we'd be setting on an existing database would be lines 91-93.

I'm open to whatever you suggest we do with this. No real preference.

scripts/run.sh
dsifford: I debated this as well, but I figured having a plugin fail to install wouldn't be enough to warrant a complete build failure.

One such example of this is if you install plugins using urls. wp-cli has no way of understanding whether or not a plugin downloaded from a url is installed (`wp plugin is-installed` doesn't accept urls). So, after the first install, it'll fail every single time (I use the term "fail" here loosely, because really, wp-cli _thinks_ it's failing, when really the plugin is just installed already).

scripts/run.sh
dsifford: Debated that as well and I'm still totally open to input. My rationale is, I think people want to explicitly know when they need to upgrade plugins. And really, the only way they'll know if an upgrade happens is if they run `docker compose logs wordpress` on every build.

Having said that, I'm not totally set on that. I'll leave that up to you.

scripts/run.sh
dsifford: Good idea. I'll add that in.

As of right now, the build completes with the latest default theme and zero plugins. Do you suggest we keep that as-is and then allow people to add additional themes using the `THEMES` environment variable?

karellm: Well I would set create a `THEME=twentysixteen` variable in the `docker-compose.yml`. It leaves the option to the user to pick which theme they want by default. It might then let the user pick a totally different public theme. If the user pick a theme that can't be installed we default to the latest default theme.

scripts/run.sh
karellm: Couldn't this be an ENV variable that default to the value you have here if not set?

scripts/run.sh
karellm: I would either move L90-94 to L66 (aka showing all the time at setup) or remove it altogether (if we use this in production, logging the password is not ideal). I would probably remove it since it doesn't serve any purpose, we set the wp-config.php for you and you can see these in it.

scripts/run.sh
karellm: I agree that explicit dependency update is a better pattern. But we are not providing a way to define which version you need, so on installation, it always get the latest one. That's why it feels natural to always keep the stack up to date at the latest version. That said, I think we can keep this discussion for another PR.

scripts/run.sh
dsifford: Ah, good idea. Default to twentysixteen if nothing set but allow overrides if set. Brilliant. Love it.

scripts/run.sh
karellm: We could change the logic:

```
if already_installed
  echo "already installed"
else if install ok
  echo "install ok"
else
  echo "error installing"
end
```

wdyt?

scripts/run.sh
dsifford: Fair point.. I think adding a version dependency system similar to npm wouldn't be terribly difficult... So for example, `PLUGINS=plugin-one@1.3.1,plugin-two@3.5.2` etc..

karellm: I wasn't suggesting we do that. It is wordpress we are talking about :smile: 

But actually, it would be simple relying on `wp-cli`, there is a [version option](http://wp-cli.org/commands/plugin/install/).

scripts/run.sh
dsifford: Not that I know of. I'm not sure how we'd be able to save something this intricate as an environment variable without saving each line as its own variable.

scripts/run.sh
dsifford: Cool, I'm good with axing this altogether as well

scripts/run.sh
dsifford: Not sure if I understand the difference. Can you explain?

karellm: Basically just to log there was an error installing the plugin, without trigger a build failure. But the logs give a sense it installed even if it didn't. Or maybe I'm missing something here :)

scripts/run.sh
dsifford: Just be sure to write somewhere in the readme that the default username and pass are both 'root'

scripts/run.sh
dsifford: Oh, gotcha. That log already exists natively with wp-cli. 

Here's an example...
```sh
wordpress_1 | => Checking plugins...
wordpress_1 | => (1/9) Plugin 'google-analytics-dashboard-for-wp' found. SKIPPING...
wordpress_1 | => (2/9) Plugin 'simple-tooltips' found. SKIPPING...
wordpress_1 | => (3/9) Plugin 'query-monitor' found. SKIPPING...
wordpress_1 | => (4/9) Plugin 'tablepress' found. SKIPPING...
wordpress_1 | => (5/9) Plugin 'ultimate-member' found. SKIPPING...
wordpress_1 | => (6/9) Plugin 'wp-mail-smtp' found. SKIPPING...
wordpress_1 | => (7/9) Plugin 'wpdiscuz' found. SKIPPING...
wordpress_1 | => (8/9) Plugin 'http://bit.ly/um-recaptcha' not found. Installing...
wordpress_1 | Downloading install package from http://bit.ly/um-recaptcha...
wordpress_1 | Unpacking the package...
wordpress_1 | Installing the plugin...
wordpress_1 | Plugin install failed.
wordpress_1 | => (9/9) Plugin 'http://bit.ly/um-online' not found. Installing...
wordpress_1 | Downloading install package from http://bit.ly/um-online...
wordpress_1 | Unpacking the package...
wordpress_1 | Installing the plugin...
wordpress_1 | Plugin install failed.
wordpress_1 |   =======================================
wordpress_1 |      WordPress Configuration Complete!
wordpress_1 |   =======================================
```

scripts/run.sh
karellm: I mean the first time your run `docker-compose up`, it will generate the `wp-cli.yml` file with it. I didn't mean something dynamically updating.

scripts/run.sh
karellm: :+1:

scripts/run.sh
karellm: Oh great, then discard my comment :+1:

scripts/run.sh
dsifford: That's what it's doing now. The other option I suppose would be just to include it as a separate file

karellm: I think I'm not being clear. I only suggest we do something like:

```
[ "$ADMIN_EMAIL" ] || ADMIN_EMAIL="admin@${DB_NAME}.com"
```

and then:

```
admin_email: ${ADMIN_EMAIL}
```

dsifford:
@karellm About to push the changes now... 

Changes include:
- Fix spacing
- Add option to specify default themes (business logic on lines 137-160).
- Unindented wp-cli.yml
- Removed unnecessary items from dockerfile and scripts folder + moved run.sh to top level
- Remove "Default Credentials" logging

Let me know if I missed anything :+1:

dsifford:
Boo for conflicts. :-1:

scripts/run.sh
dsifford: Ohh, ok I see.. let me fix that real fast and push over the changes. Just a sec..

dsifford:
Note to self: this build will fail at present because the MySQL ping loop is using the host name of my fork and not of this repo. 

Will change later today...

dsifford:
@karellm Before I go about fixing the issue I mentioned above, what are your thoughts on switching the MySQL container's hostname from `mysql` to `db` -- The rationale here being, it should be possible for people to also use MariaDB with this build as well.. Let me know!

Otherwise, lines 26 and 40 need to be fixed.

karellm:
I like the change! Let's use db

dsifford:
Sounds good!

dsifford:
@karellm Just made two changes:

First, the Dockerfile for some reason was missing the `wp core download` and `chown /app ...etc`  lines (likely my fault), so I added those back.

I also moved the 'ENV DEBIAN_FRONTEND noninteractive` inline on the `apt-get` install line. This is because using noninteractive mode unsets `TERM`, and if people want to use mariadb, they need `TERM` set (tested: fixed).

Last, I added your name and email to the top as the `MAINTAINER`.

Let me know if I forgot anything.

karellm:
Sounds good. Thanks again for the great work!. For the merge conflict can you rebase off of master?

dsifford:
@karellm Scratch my last. I think the issue I was facing may have been the result of the dismal network conditions I had where I was working. Back on my main machine now and it looks like both mariadb and mysql work as expected. 

Also, rebase: done.

karellm:
Great, thanks!

PR #18, opened 2016-03-12
Update the README

karellm:
@dsifford Still a WIP but it updates the documentation. 

- Update the documentation
  - [x] Readme
  - [ ] Blog
- Small formatting and message fixes to `run.sh`:
  - [x] Fix indentations and limit line to a maximum of 80 char as much as possible
  - [x] Make code blocks more consistent (remove spaces and add visual sections)

1/ Importing and exporting the DB could use the `SEARCH_REPLACE` feature, making the command to run long and ugly. It might be worth creating an alias somewhere (or script file).

2/ You didn't keep the gulp part. Any reason for that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,33 +1,124 @@
 # Visible Wordpress Starter
 
-A project to make Wordpress development slightly less frustrating created by the team at [Visible](https://visible.vc/).
+A Docker Wordpress development environment by the team at [Visible](https://visible.vc/) and some awesome [contributors](https://github.com/visiblevc/wordpress-starter/graphs/contributors). Our goal is to make Wordpress development slightly less frustrating.
 
 ## Requirements
 
-- Docker
-- Docker Compose 1.6+
+Well, to run a Docker environment, you will need Docker. The Dockerfile is only for an Apache+PHP+Wordpress container, you will need a MySQL container to run a website. We use Docker Compose 1.6+ for the orchestration.
+
+---
 
 ## Getting started
 
+This project has 2 parts: the Docker environment and a set of tools for theme development. To quickly get started, you can simply run the following:
+
 ```
-# close this project
+# copy the files
 git clone https://github.com/visiblevc/wordpress-starter.git
-# remove the .git folder and Dockerfile (we use the docker hub image)
-rm -rf .git Dockerfile
-# start the containers
+rm -rf .git Dockerfile run.sh
+
+# start the website at localhost:8080
 docker-compose up
-# visit localhost:8080
 ```
 
-## Documentation
+---
 
-We wrote a series of articles to document the project:
+### Documentation
+
+We wrote a series of articles explaining in depth the philosophy behind this project:
 
 - [Intro: A slightly less shitty WordPress developer workflow](https://visible.vc/engineering/wordpress-developer-workflow/)
 - [Part 1: Setup a local development environment for WordPress with Docker](https://visible.vc/engineering/docker-environment-for-wordpress/)
 - [Part 2: Setup an asset pipeline for WordPress theme development](https://visible.vc/engineering/asset-pipeline-for-wordpress-theme-development/)
 - [Part 3: Optimize your wordpress theme assets and deploy to S3](https://visible.vc/engineering/optimize-wordpress-theme-assets-and-deploy-to-s3-cloudfront/)
 - Part 4: Auto deploy your site on your server (coming)
+
+### The Docker environment
+
+The only thing you need to get started is a `docker-compose.yml` file:
+
+```yml
+version: '2'
+services:
+  data:
+    image: busybox
+    volumes:
+      - /app
+      - /var/lib/mysql
+  wordpress:
+    image: visiblevc/wordpress:latest
+    links:
+      - db
+    ports:
+      - 8080:80
+      - 443:443
+    volumes:
+      - ./data:/data # Required if importing an existing database
+      - ./wp-content/uploads:/app/wp-content/uploads
+      - ./yourplugin:/app/wp-content/plugins/yourplugin # Plugin development
+      - ./yourtheme:/app/wp-content/themes/yourtheme   # Theme development
+    environment:
+      DB_NAME: wordpress
+      DB_PASS: root # must match below
+      PLUGINS: academic-bloggers-toolkit,co-authors-plus
+      SEARCH_REPLACE: yoursite.com,localhost:8080
+  db:
+    image: mysql:5.7
+    ports:
+      - 3306:3306
+    volumes:
+      - data:/var/lib/mysql
+    environment:
+      MYSQL_ROOT_PASSWORD: root
+volumes:
+  data: {}
+```
+
+##### MySQL Credentials
+
+- username: `root`
+- password: `root` (can be changed with the `MYSQL_ROOT_PASSWORD` and `DB_PASS` environment variables)
+- database: `wordpress` (can be changed with the `DB_NAME` environment variable)
+- admin email: `admin@${DB_NAME}.com`
+
+##### WordPress Container Environment variables
+
+- `DB_PASS` (required): Must match `MYSQL_ROOT_PASSWORD` of the mysql container
+- `DB_NAME` (optional): Defaults to `wordpress`
+- `ADMIN_EMAIL` (optional): Defaults to `admin@${DB_NAME}.com`
+- `THEMES` (optional): Comma-separated list of themes you want to install.
+- `PLUGINS` (optional): Comma-separated list of plugins you want to install.
+- `SEARCH_REPLACE` (optional): Comma-separated string in the form of `current-url,replacement-url`.
+    - When defined, `current-url` will be replaced with `replacement-url` on build (useful for development environments utilizing a database copied from a live site).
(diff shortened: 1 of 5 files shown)


dsifford:
We use gulp for css/js minification / autoprefixing, but I've found that gulp is sort of a nightmare when it comes to actually doing shell-related tasks.

We currently use a a file that I called `syncup.sh` to handle the various build tasks not handled with the dockerfile (see below). Another option for the less shell-saavy would be to use npm run commands for those types of things. Example: `npm run get-database` could run a giant shell command separated by `&&`.

Here's our `syncup.sh` file

```sh
#!/bin/bash

###
# COMANDS
# 'get uploads' = retrieve all uploads that do not already exist
# 'get plugin [plugin-name]' = retrieve and overwrite the plugin [plugin-name]
# 'get --all-plugins' = retrieve all paid plugins (REQUIRED BEFORE RUNNING ENVIRONMENT)
# 'get database' = retrieve a copy of the database from the server + rename urls
# 'update plugin [plugin-name] = update a specific plugin'

# Globals
SCRIPTDIR=$(dirname "$0")
PaidPlugins=(learndash-propanel sfwd-lms um-profile-completeness)

# OS-specific variables
if [[ $(uname) == "Linux" ]]; then
  ONLINUX=true
  CONTAINER_IP="localhost"
else
  ONLINUX=false
  CONTAINER_IP=$(docker-machine ip "$DOCKER_MACHINE_NAME")
fi

cd "$SCRIPTDIR" || exit

case "$1" in

    get)

        case "$2" in

            uploads)

                cd "$SCRIPTDIR/wp-content" || exit
                echo "=> Downloading uploads from server..."
                wget -rq -nH --cut-dirs=1 -l inf --no-clobber ftp://ftp.SITEURL.com/wp-content/uploads/ --ftp-user=<USERNAME-HERE> --ftp-password=<PASSWORD-HERE>
                echo "=> Uploads Synced Successfully!"
                ;;

            --all-plugins)

                cd "$SCRIPTDIR/wp-content/plugins" || exit
                for plugin in "${PaidPlugins[@]}"; do
                    echo "=> Retrieving plugin: $plugin"
                    wget -m -nH --cut-dirs=2 ftp://ftp.SITEURL.com/wp-content/plugins/"$plugin" --ftp-user=<USERNAME-HERE> --ftp-password=<PASSWORD-HERE>
                    echo "=> Plugin $plugin Retrieved Successfully!"
                done
                ;;

            plugin)

                cd "$SCRIPTDIR/wp-content/plugins" || exit
                echo "=> Retrieving plugin: $3"
                wget -m -nH --cut-dirs=2 ftp://ftp.SITEURL.com/wp-content/plugins/"$3" --ftp-user=<USERNAME-HERE> --ftp-password=<PASSWORD-HERE>
                echo "=> Plugin $3 Retrieved Successfully!"
                ;;

            database)

                # Ensure we are in the script's directory
                cd "$SCRIPTDIR/data" || exit

                # Ensure proper file permissions
                chmod 400 aliemu_dsa

                # Run ssh-agent and add the ssh private key
                eval "$(ssh-agent -s)"
                ssh-add aliemu_dsa

                # SSH into siteground and backup the database and copy to the data directory
                if [[ $ONLINUX == true ]]; then
                    ssh -i aliemu_dsa sshuser@sshsite.net -p 18765 "
                    cd public_html
                    wp db export database.sql"
                    scp -i aliemu_dsa -P 18765 sshuser@sshsite.net:public_html/database.sql .
                else
                    ssh -i aliemu_dsa sshuser@sshsite.net -p 18765 -o "PubkeyAcceptedKeyTypes ssh-dss" "
                    cd public_html
                    wp db export database.sql"
                    scp -i aliemu_dsa -P 18765 -o "PubkeyAcceptedKeyTypes ssh-dss" sshuser@sshsite.net:public_html/database.sql .

                    # Machine not running? Start it up!
                    if [[ $(docker-machine status "$DOCKER_MACHINE_NAME") != 'Running' ]]; then
                        cd "$SCRIPTDIR" || exit
                        docker-compose up -d
                    fi
                fi

                # Import the database
                echo "=> Importing database..."
                docker exec -it "$(docker ps -lq)" bash -c "wp db import /data/database.sql --allow-root"

                # Replace live URL with dev URL
                echo "=> Replacing URLs..."
                docker exec -it "$(docker ps -lq)" bash -c "wp search-replace  'http://www.SITEURL.com' 'http://${CONTAINER_IP}:8080' --skip-columns=guid --allow-root"
                echo "=> Database Successfully Imported!"
                ;;

            *)
                echo "'get' subcommmand must be either 'uploads', 'plugin', or 'database'"
                exit 1
        esac
        ;;

    update)

        case "$2" in

            plugin)

                echo "=> Updating plugin: $3"
                PLUGIN=$3
                docker exec -it "$(docker ps -lq)" bash -c "wp plugin install $PLUGIN --force --activate --allow-root"
                ;;

            *)
                echo "'update' subcommand must be 'plugin'"
                exit 1
        esac
        ;;

    *)
        echo "The only commands available are 'get' and 'update'"
        exit 1
esac
```

From the root directory of our project, running `./syncup.sh get database` grabs a new copy of the site's database, uploads it to the container, and replaces the URLs.

karellm:
Interesting, I actually don't like gulp (or build tools) but used it to move fast since I knew it well. Using npm could definitely be a good alternative. I will review your script and consider options.

karellm:
@dsifford If made a cleanup pass and realized I can run `docker-compose up` with the new `run.sh`. It complains about `tput: No value for $TERM and no -T specified` and all commands fail. I need to investigate that more.

Let me know what you think of the changes too.

Finally, couple things I noticed:

1. Plugins are installed every time, themes not. Also relying on checking the folder akismet is not reliable since someone might want to use that plugin. Maybe `hello` would be a better pick.
2. We might want a script to run commands like backup/restore the DB. I like your `syncup.sh` and we might be able to reuse that

dsifford:
@karellm Is the branch you are running from using the latest version of `run.sh`? I believe I added the `-T` flags to the tput commands.

If so, we can just remove the prettiness altogether if tput is going to cause a problem.

For themes being installed every time: The concern I had there was if someone made changes to the theme's style or options. Would those get overwritten every time the environment is ran?

Re scripts: I can write up a started `package.json` if using npm run commands was something that perhaps interested you. Otherwise, open to suggestions.

karellm:
So the `-T` helped but didn't solve the underlying error. If I run `sudo -u www-data wp core config` inside a `docker-compose run wordpress bash`, I get `PHP Warning:  file_put_contents(/app/wp-config.php): failed to open stream: Permission denied in phar:///usr/local/bin/wp/php/commands/core.php on line 347`.

It is clearly a file/folder permission issue. Do you have any idea why?

dsifford:
That happened to me as well! That's the strange issue that I was talking about that happened when I wasn't at my usual computer. I can't reproduce the issue from here.

The issue is: For some reason, wordpress core isn't being downloaded into the `/app` directory. So everything from then on fails (because the files don't exist).

dsifford:
if you run `docker exec -it $(docker ps -lq) /bin/bash` and then `ls` you'll see the `/app` directory is empty. Not sure why.

dsifford:
If you run `wp core download` before `wp core config` everything will work.

dsifford:
I suppose we could try moving the download line into the `run.sh` script and see if that helps?

dsifford:
Actually, standby for a bit.. I think we'd be better off grabbing wordpress manually using curl in the Dockerfile. Give me a few and I'll add that to my current PR.

karellm:
I do have the files, I don't think it is the issue. The issue is the permissions on the folder I think.

That said, I believe downloading wordpress should be the role of the `run.sh` rather than the image. Otherwise the image will potentially get an outdated version of wordpress. And I don't plan on deploying a new one at every wordpress release.

karellm:
So mounting the `data:/app` line in the docker file is messing with the folder permissions.

```
# without the line
drwxr-xr-x   6 www-data www-data 4096 Mar 17 19:16 app

# with the line
drwxr-xr-x   6  999   999 4096 Mar 17 19:15 app
```

dsifford:
@karellm Why is data being mounted into app? That's a typo. Should be mounted into `/data`

Also, check my pull. I moved the wordpress install into run.sh

karellm:
My initial intention was to persist file changes across containers but this shouldn't be needed.

dsifford:
I see. Yeah, not needed. By creating a volume, you'll persist changes, no matter where the location might be.

dsifford:
I guess I should also mention, if you aren't trashing your compose builds that have volumes by using `docker-compose down -v` then your `/var` directory is probably jam-packed with old volumes. 

(I made that mistake in the past and it took up a huge amount of space)

Same goes for `docker rm -v <image-ID>` :+1:

karellm:
I have a bunch of routines to stop and rm containers and volumes. Thanks for the tip :+1: 

@dsifford See the latest change in this PR, all should be good. If you are happy with the changes, let's merge!

README.md
```diff
@@ -29,6 +33,94 @@ We wrote a series of articles to document the project:
...
+      - /var/lib/mysql
+  wordpress:
+    image: visiblevc/wordpress:latest
+    links:
+      - db
+    ports:
+      - 8080:80
+      - 443:443
+    volumes:
+      - data:/app
```
dsifford: Double check this to make sure it matches the current `docker-compose.yml` file

dsifford:
@karellm Other than my comment, everything looks great! Nice work :+1:

PR #20, opened 2016-03-17
WIP: Package.json commands

dsifford:
~~This is a WIP for various utility commands built into package.json.~~

~~Needed:~~
- [x] ~~Export MySQL dump~~
- [x] ~~Import MySQL dump~~
- [x] ~~Update plugin~~
- [x] ~~Update all plugins~~
- [x] ~~Install plugin~~
- [x] ~~Update theme~~
- [x] ~~Update all themes~~

**Edit: (03/20/16)**
This PR now contains one single npm run script which allows near-native use of wp-cli as well as a change in `run.sh` which causes `wp-cli.yml` and `wp-config.php` to be recreated on each build so that any changes to the associated environment variables are always applied after the initial build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,15 +2,7 @@
   "name": "visible-wordpress",
   "version": "1.0.0",
   "scripts": {
-    "export-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db export /data/database.sql\"",
-    "import-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db import /data/database.sql\" && npm run search-replace",
-    "update-plugins": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update --all\"",
-    "update-themes": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update --all\"",
-    "plugin-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin install $PLUGIN\"",
-    "plugin-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update $PLUGIN\"",
-    "theme-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update $THEME\"",
-    "theme-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme install $THEME\"",
-    "search-replace": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp search-replace 'http://www.oldurl.com' 'http://localhost:8080' --skip-columns=guid\""
+    "wp": "wpcmd() { docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp $(echo $@)\"; };wpcmd "
   },
   "dependencies": {},
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,17 @@
 {
-  "name": "visible-website",
+  "name": "visible-wordpress",
   "version": "1.0.0",
+  "scripts": {
+    "export-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db export /data/database.sql\"",
+    "import-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db import /data/database.sql\" && npm run search-replace",
+    "update-plugins": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update --all\"",
+    "update-themes": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update --all\"",
+    "plugin-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin install $PLUGIN\"",
+    "plugin-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update $PLUGIN\"",
+    "theme-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update $THEME\"",
+    "theme-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme install $THEME\"",
+    "search-replace": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp search-replace 'http://www.oldurl.com' 'http://localhost:8080' --skip-columns=guid\""
+  },
   "dependencies": {},
   "devDependencies": {
     "chalk": "^1.1.0",
--- a/run.sh
+++ b/run.sh
@@ -15,7 +15,6 @@ ERROR () {
   exit 1;
 }
 
-
 # Configure wp-cli
 # ----------------
 if [ ! -f /app/wp-cli.yml ]; then
--- a/run.sh
+++ b/run.sh
@@ -17,7 +17,6 @@ ERROR () {
 
 # Configure wp-cli
 # ----------------
-if [ ! -f /app/wp-cli.yml ]; then
 cat > /app/wp-cli.yml <<EOF
 quiet: true
 apache_modules:
@@ -39,7 +38,6 @@ core install:
   admin_email: $ADMIN_EMAIL
   skip-email: true
 EOF
-fi
 
 
 # Download WordPress
@@ -70,17 +68,10 @@ printf "\t%s\n" \
 # wp-config.php
 # -------------
 printf "=> Generating wp.config.php file... "
-if [ ! -f /app/wp-config.php ]; then
-  sudo -u www-data wp core config >/dev/null 2>&1 || \
-    ERROR $LINENO "Could not generate wp-config.php file"
-  printf "Done!\n"
-else
-  printf "Already exists!\n"
-fi
-
-# Enable or Disable debug mode
-sed -i "s/define('WP_DEBUG', \(true\|false\))/define('WP_DEBUG', ${WP_DEBUG,,})/g" /app/wp-config.php
-
+rm -f /app/wp-config.php
+sudo -u www-data wp core config >/dev/null 2>&1 || \
+  ERROR $LINENO "Could not generate wp-config.php file"
+printf "Done!\n"
 
 # Setup database
 # --------------


dsifford:
@karellm Check these out when you get a moment.

There just one slight proviso when using these that you should be aware of (I'll comment on those directly)...

package.json
```diff
@@ -1,6 +1,16 @@
...
-  "name": "visible-website",
+  "name": "visible-wordpress",
   "version": "1.0.0",
+  "scripts": {
+    "export-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db export /data/database.sql\"",
+    "import-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db import /data/database.sql\"",
+    "update-plugins": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update --all\"",
+    "update-themes": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update --all\"",
+    "plugin-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin install $PLUGIN\"",
+    "plugin-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update $PLUGIN\"",
```
dsifford: For lines 9-10 to work correctly, they need to be ran like this:

(Assuming that `some-plugin` is the one we're interested in installing)
`PLUGIN=some-plugin npm run plugin-install`

package.json
```diff
@@ -1,6 +1,16 @@
...
+    "theme-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update $THEME\"",
+    "theme-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme install $THEME\""
```
dsifford: Lines 11-12 need to be ran the same as above, only using `THEME`...

`THEME=some-theme npm run theme-install`

run.sh
```diff
@@ -10,7 +10,14 @@
 || SEARCH_REPLACE=false
 
 # Error handling
-ERROR () { echo -e "\n=> $(tput setaf 1)$(tput bold)ERROR$(tput sgr 0) (Line $1): $2."; exit 1; }
+ERROR () { echo -e "\n=> $(tput -T xterm setaf 1)$(tput -T xterm bold)ERROR$(tput -T xterm sgr 0) (Line $1): $2."; exit 1; }
+
+# Download WordPress
+curl -O https://wordpress.org/latest.zip \
```
dsifford: Tbh, this whole stretch of code can probably be replaced with wp-cli

package.json
```diff
@@ -1,6 +1,17 @@
...
+  "scripts": {
+    "export-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db export /data/database.sql\"",
+    "import-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db import /data/database.sql\" && npm run search-replace",
+    "update-plugins": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update --all\"",
+    "update-themes": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update --all\"",
+    "plugin-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin install $PLUGIN\"",
+    "plugin-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update $PLUGIN\"",
+    "theme-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update $THEME\"",
+    "theme-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme install $THEME\"",
+    "search-replace": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp search-replace 'http://www.oldurl.com' 'http://localhost:8080' --skip-columns=guid\""
```
dsifford: Users will have to change the two URLs here according to their own needs

karellm:
@dsifford Can you rebase from `master`? Wordpress download is already fixed.

dsifford:
@karellm Done!

dsifford:
That said, I'm still not totally in love with this.. If you want, we can let this linger for a bit longer and I'll brainstorm better options...

run.sh
```diff
@@ -70,17 +68,10 @@ printf "\t%s\n" \
...
-    ERROR $LINENO "Could not generate wp-config.php file"
-  printf "Done!\n"
-else
-  printf "Already exists!\n"
-fi
-
-# Enable or Disable debug mode
-sed -i "s/define('WP_DEBUG', \(true\|false\))/define('WP_DEBUG', ${WP_DEBUG,,})/g" /app/wp-config.php
-
+rm -f /app/wp-config.php
```
dsifford: @karellm Thoughts on this (and line 20)?

This way, `wp-cli.yml` and `wp-config.php` are always overwritten (and should always contain the right options) on each build.

package.json
```diff
@@ -2,15 +2,7 @@
...
-    "import-db": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp db import /data/database.sql\" && npm run search-replace",
-    "update-plugins": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update --all\"",
-    "update-themes": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update --all\"",
-    "plugin-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin install $PLUGIN\"",
-    "plugin-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp plugin update $PLUGIN\"",
-    "theme-update": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme update $THEME\"",
-    "theme-install": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp theme install $THEME\"",
-    "search-replace": "docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp search-replace 'http://www.oldurl.com' 'http://localhost:8080' --skip-columns=guid\""
+    "wp": "wpcmd() { docker exec $(docker ps -lq) /bin/bash -c \"sudo -u www-data wp $(echo $@)\"; };wpcmd "
```
dsifford: @karellm Bingo. This is the winner... Using this new run script, wp-cli feels like it's running natively but runs within the wordpress container.

So, for example...

- `npm run wp plugin install <some-plugin>` installs `<some-plugin>` inside the container
- `npm run wp db import /data/database.sql` imports a the database
...etc, etc, etc.


If you're cool with this and the last change, let me know and I'll rebase. :+1:

dsifford:
@karellm Are we able to merge this? Anything you need addressed?

run.sh
karellm: :+1:

karellm:
I just got busy, looks great thanks!

package.json
joggienl: Hi!

Very nice solution having one command to use the wp-cli! Very helpfull in the development workflow! I do have some problems with this sollution though. Number one: it does not work on Windows.

There are two problems regarding to windows:
- You can't define functions like that on the cmd prompt
- You can't substitute a command's output like with $(docker ps -lq).

The only possible solution I can think of right now is to make a script or separate npm module that does this calling.

Second problem I found with this method is that you can't be sure that `docker ps -lq` outputs the right last made container. For example, if you run the `docker-compose up` command, play around a bit. And after that run another docker command like `docker run hello-world`, the docker ps -lq outputs the last container made: the hello-world container.

You can also run `docker-compose ps` from the folder with the docker-compose.yml file. This command seems to output only the containers that are defined from the yml file. However you still need a way to verify the right container to send the command to. You could do some smart guessing, docker-compose takes the folder name of the project + the service name + a sequence number as the container name. Maybe if you could point to the WordPress container only and send the command there, your done. Even if you would have more WordPress containers (in a scaled environment) this would most probably not be a problem because the database would be separate (on the other hand this would only be the case in production. Are you going to use wp-cli there? Preferably not!).

You could use the container_name directive in the .yml file. This way you would be certain about the container its name and use that safely. This is considered a bad practice: docker requires containers to have unique names and this would get in your way when you would need to scale.

That made me think. Maybe it's perfectly legit to define a separate service in the .yml file. a named container just for the wp-cli tool. For development that would be perfectly fine, just don't push it to production.

I really like this being part of the wordpress-starter project. I do feel the need to say that it must be more a more stable solution than it is now. Otherwise it would be better to document the possibility to check the container name for yourself (as a developer) and then run the docker exec command for yourself.

joggienl:
I've added a note to this issue: https://github.com/visiblevc/wordpress-starter/pull/20#discussion_r60827635

Curious what your thoughts are about this!

karellm:
@joggienl Thanks for great feedback. I indeed ran into the `docker ps -lq` issue. If I recall properly, you could use `docker-compose run wordpress wp-cli ...`. Have you tried that?

Also I don't have access to a windows machine, nor any interest setting one up. If you can come up with a solution, a PR would be greatly appreciated!

joggienl:
If possible, that would be great!

I've tried and failed so far... by any change know what this error means?

```
D:\Path\> docker-compose exec wordpress wp theme list
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "compose\cli\main.py", line 57, in main
  File "compose\cli\main.py", line 108, in perform_command
  File "compose\cli\main.py", line 347, in exec_command
NameError: global name 'ExecOperation' is not defined
docker-compose returned -1
```

dsifford:
@joggienl `exec` isn't a compose command. You're looking for `run`.

joggienl:
@dsifford it is actually: https://docs.docker.com/compose/reference/exec/

CLI outputs that it is ment to "Execute a command in a running container". That's exactly what we want?

The `run` command will start up a new container, at least that is wat happens on windows... and it won't work as expected.. Maybe OSX gives different output? Not behind my mac today... This is what windows says:

```
D:\Path\> docker-compose run -d -T wordpress wp theme list --allow-root
path_wordpress_run_5

D:\Path\> docker logs krachtdoorlerennl_wordpress_run_5
Error: This does not seem to be a WordPress install.
Pass --path=`path/to/wordpress` or run `wp core download`.
```

As you can see I need to run de logs command separately. This is because some missing features in docker-compose for windows. Maybe OSX does it as expected at this point.

```
D:\Path\> docker-compose run wordpress wp theme list --allow-root
ERROR: Interactive mode is not yet supported on Windows.
Please pass the -d flag when using `docker-compose run`.
```

Either way, I don't think that you want to "boot up" a complete container just to fire some wp-cli calls.

joggienl:
It feels like there could be some thoughts that go like "why windows!?" ;-)

dsifford:
@joggienl 

On linux: 

```sh
$ docker-compose --version
docker-compose version 1.6.2, build unknown
$ docker-compose exec
No such command: exec
```
The `exec` command that I'm familar with is a Docker Engine command. `docker exec -it <some-hash> /bin/bash <some command>` executes a command in a single running container.

There are separate commands for compose however which allows you to refer to the containers by name, rather than by hash.

Also, just now reading your message. Why does the npm command not work on windows? Are you not using cygwin/conemu/msys2? If not, I'd say that's a much more prudent issue than the npm command.

Re: `docker ps -lq` --- I'm never working on more than one thing at a time, but I suppose if you're a multitasker, that might be an issue. 

The best part about this orchestration is that it's fully disposable. When I'm done for the day, I remove it completely. The next day, I start fresh (build time from scratch takes me ~2 minutes). That way I'm guaranteed to have a database that reflects the most current database in the live site and a clean slate to work from. Additionally, this prevents stray volumes from accumulating in your `/var` directory (or wherever they are stored on windows)

dsifford:
Update: They must have added `exec` to compose 1.7.0 👍 

```sh
$ dc --version
docker-compose version 1.7.0, build 0d7bf73
dsifford ~
$ dc exec
Execute a command in a running container

Usage: exec [options] SERVICE COMMAND [ARGS...]

Options:
    -d                Detached mode: Run command in the background.
    --privileged      Give extended privileges to the process.
    --user USER       Run the command as this user.
    -T                Disable pseudo-tty allocation. By default `docker-compose exec`
                      allocates a TTY.
    --index=index     index of the container if there are multiple
                      instances of a service [default: 1]
```

joggienl:
@dsifford 

```
docker-compose version
docker-compose version 1.7.0, build 0d7bf73
docker-py version: 1.8.0
CPython version: 2.7.11
OpenSSL version: OpenSSL 1.0.2d 9 Jul 2015
```

The `docker-compose exec` is new indeed. It's in the changelog for 1.7.0.
Does the exec command work for you on Linux?

I am using cmder, that's conemu plus some more packages arround it. Otherwise it would be pretty hard to do CLI tasks anyway on windows... substituting commands with $(command) or things like eval is not possible. Also defining functions like `func () { do_stuff; }; func` is not possible. Maybe when Microsoft releases their bash implementation it is working on, it can be done.

I can follow your argument about the multitasking and doing one thing at a time. Still it is way to easy to get it in a non working state ("broken" is not the right word here). I would want to be sure (as most as I can) that this command always works, regardless what I do in between. Also if this would be part of a CI environment (it could...) than you would not like to configure things and then say "well, if another container would get fired-up in between, this won't work".

dsifford:
@joggienl Fair enough.

I'm an ex-cmder user myself. I was a windows guy up until about a year ago and I got fed up, went to linux, and never looked back.

But anyway, I'm digging the new `exec` command. It looks like in order to have it work for wp-cli though you'll probably have to give it the `--priveledged` flag (since it uses sudo) 

You can modify the current command to this though and it'll probably work...

`docker-compose exec --user www-data wordpress  wp theme list`

let me know if that works for you.

joggienl:
@dsifford 

Does the command work for you on Linux? On Windows, whatever command I try to pass to a service, I alway get the message as noted above. I've created an issue at docker/compose. I'm guessing something isn't allright yet for the windows build regarding this. See it here: https://github.com/docker/compose/issues/3374

And off-topic about the "windows guy", I've always used Windows/Linux/OSX. I do have my preference (off course) though but also like to work with all of them. On Linux and OSX I always liked bash. Cmder does some great things regarding terminal/prompt tools, but in the end it does not compare to bash (yet).

dsifford:
@joggienl Yeah, it works for me. Just to confirm though, in your compose file, your wordpress service is indeed named "wordpress" correct? If not, you'll have to use that name as the service name for your command.

joggienl:
@dsifford I can confirm it is still named "wordpress" :+1: I've also tried on another container build and brought up with composer (a project nothing to do with this wordpress-starter). I'm getting the same error over there. That's also why I've created the issue on the docker/compose repo.

dsifford:
Dang it, Windows! 😂

PR #21, opened 2016-03-17
Add an option for WP_DEBUG

karellm:
A quick improvement that let you set `WP_CONFIG`. I needed that while developing a theme

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-[ "$DB_NAME" ] || DB_NAME='wordpress'
-[ "$DB_PASS" ] || DB_PASS='root'
-[ "$THEMES" ]  || THEMES='twentysixteen'
+[ "$DB_NAME" ]  || DB_NAME='wordpress'
+[ "$DB_PASS" ]  || DB_PASS='root'
+[ "$THEMES" ]   || THEMES='twentysixteen'
+[ "$WP_DEBUG" ] || WP_DEBUG='false'
 [ "$ADMIN_EMAIL" ] || ADMIN_EMAIL="admin@${DB_NAME}.com"
 [ "$SEARCH_REPLACE" ] && \
   BEFORE_URL=$(echo "$SEARCH_REPLACE" | cut -d ',' -f 1) && \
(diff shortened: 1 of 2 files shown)


karellm:
@dsifford Can you review this? Thanks

dsifford:
@karellm Looks alright. 

What about doing it like this instead? https://github.com/wp-cli/wp-cli/blob/4952290c9b6d14d938ac8f9fbf3f9439df83573b/php/commands/core.php#L324

Your call. I'll be able to test tomorrow.

karellm:
What's the benefit? I was thinking you might want to switch it on and off. And `wp core config` would work only once (though we can remove and regenerate it).

dsifford:
No benefit. Didn't know it was something u wanted switched on and off.

karellm:
@dsifford Ok let me know if we are good to merge!

run.sh
```diff
@@ -76,6 +79,9 @@ else
   printf "Already exists!\n"
 fi
 
+# Enable or Disable debug mode
```
dsifford: What's the purpose of this?

run.sh
```diff
@@ -28,6 +29,8 @@ core config:
   dbpass: $DB_PASS
   dbname: $DB_NAME
   dbhost: db:3306
+  extra-php: |
+    define('WP_DEBUG', ${WP_DEBUG,,});
```
dsifford: Nothing wrong here, just a question for ya...

What does the `,,` do in `${WP_DEBUG,,}`? I've never seen that before. Can you explain? :smile:

karellm: It returns the lowercase version of `WP_DEBUG`. For some reason docker-compose would capitablize the environment variable.

run.sh
karellm: `wp-cli.yml` is generated only once. We need this to subsequently change it.

dsifford: Not sure if I'm understanding.. Why is this needed if you're explicitly setting WP_debug with `wp-cli.yml`? If I'm reading this correctly, won't this just set the same value twice? Am I missing something?

If the goal is to generate a new `wp-config` file after build success (using a different option than was set initially), why not just write an npm run command to delete the current config file and then regenerate a new one using `wp core config [...]`?

karellm: On first run yes. But if you stop the container and start it again with the ENV variable, the `wp-cli.yml` won't get generated a second time. But I like your idea of a script I think

dsifford: Ah, ok I understand now. Makes sense I suppose.

karellm:
Merging, we can work on replacing with a script once we figure out what we want on that level.

PR #22, opened 2016-03-18
Support multiline definition for plugins

karellm:
Our website uses a lot of plugins, and we don't write code longer than 80 chars. It felt highly ineffective to list plugins in a single line. This solve that.


dsifford:
@karellm Hear, hear! :confetti_ball: :tada: 

Until today, I was totally unaware of yaml's `|` and `>` options... Totally a gamechanger. 

Strong work! :+1:

PR #26, opened 2016-04-14
also cleanup generated templates

e2:
Before, clean only removes files in dist/.
After this, files copied from assets/templates/ are also deleted during 'clean'

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -16,6 +16,8 @@ var uglify       = require('gulp-uglify');
 var minifyCss    = require('gulp-minify-css');
 var awspublish   = require('gulp-awspublish');
 var autoprefixer = require('gulp-autoprefixer');
+var glob = require('glob');
+var del = require('del');
 
 // Configuration
 var config = JSON.parse(fs.readFileSync('./gulp.json'));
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -56,7 +58,15 @@ gulp.task('deploy-assets', function (callback) {
   );
 });
 
+var generated_template_for = function(file) {
+  return file.replace(assetPath + '/templates/', themePath + '/');
+}
+
 gulp.task('clean', function () {
+  glob(assetPath + '/templates/**/*', function(err, files) {
```
karellm: I'd prefer an atomic structure with 2 tasks `clean-templates` and `clean-dist` both being called by the `clean` task.

Also is there a reason you are using `glob` vs `gulp` here? For consistency, I'd rather use gulp unless there is a good reason.

gulpfile.js
```diff
@@ -56,7 +58,15 @@ gulp.task('deploy-assets', function (callback) {
   );
 });
 
+var generated_template_for = function(file) {
```
karellm: This should be `camelCase`.

karellm:
Thanks, I added couple comments

e2:
Done, except for the glob/del usage.

Reason: I don't know how.

gulp is based on streams, and here it's about deleting specific files from an output folder. This means, it's not about using streams, but:

1. glob
2. path name transforming
3. deleting specific files

So I don't know how to elegantly solve this with gulp. gulp-tap + gulp-clean?

Anyway, the code is improved, it's split into tasks and it works. If I can improve further, let me know how specifically. (This is pretty much the first time in my life I'm using gulp or npm or JavaScript really).

e2:
https://github.com/gulpjs/gulp/blob/master/docs/recipes/delete-files-folder.md

BTW, I think the idea of "compiling template files" by replacing fixed strings referencing assets is flawed. Since it's PHP code anyway, it's best to just use a custom `asset_path()` method, as described [here](https://github.com/sindresorhus/gulp-rev/blob/master/integration.md#approach-3---php-reads-the-manifest-and-provides-asset-names) - (linked from `gulp-rev` Readme).

PR #27, opened 2016-04-14
fix broken deploy tasks (mispellings)

e2:
Also fixed the hardcoded "visible" template name.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -21,9 +21,8 @@ var autoprefixer = require('gulp-autoprefixer');
 var config = JSON.parse(fs.readFileSync('./gulp.json'));
 var themePath = 'wp-content/themes/' + config.theme;
 var distPath  = themePath + '/dist';
+var distPathAbsolute = '/' + distPath;
 var assetPath = themePath + '/assets';
-var revision = Math.floor(Date.now() / 1000);
-
 
 gulp.task(
   'default',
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -24,6 +24,9 @@ var distPath  = themePath + '/dist';
 var assetPath = themePath + '/assets';
 var revision = Math.floor(Date.now() / 1000);
 
+// CDN path replacing
+var fullDistPath = '/' + distPath;
```
karellm: I prefer `distPathAbsolute` wdyt? Also no need for the comment and extra space, you can put under line 23 or 24.

gulpfile.js
```diff
@@ -24,6 +24,9 @@ var distPath  = themePath + '/dist';
 var assetPath = themePath + '/assets';
 var revision = Math.floor(Date.now() / 1000);
```
karellm: While we are cleaning can you remove L25, this is not used anywhere. Thanks!

gulpfile.js
```diff
@@ -169,21 +171,21 @@ gulp.task('replace-versionned-assets-in-assets', function () {
     .pipe(collect({
       replaceReved: true,
       dirReplacements: {
-        '/wp-content/themes/visible/dist': config.productionAssetURL
+        fullDistPath : config.productionAssetURL
```
karellm: This won't work. In javascript this would make the key be `'fullDistPath'`, not the value of the variable. You'd want to do: 

```js
var dirReplacements = {};
dirReplacements[fullDistPath] = config.productionAssetURL;

// ...

.pipe(collect({
  dirReplacements: dirReplacements
```

gulpfile.js
```diff
@@ -169,21 +171,21 @@ gulp.task('replace-versionned-assets-in-assets', function () {
...
+gulp.task('replace-versioned-assets-in-templates', function () {
   return gulp.src([
       themePath + '/**/*.json',
       themePath + '/*.php'
     ])
     .pipe(collect({
       replaceReved: true,
       dirReplacements: {
-        '/wp-content/themes/visible/dist': config.productionAssetURL
+        fullDistPath : config.productionAssetURL
```
karellm: Same

karellm:
Thanks for the PR, I added couple comments

e2:
Done

karellm:
Looks great, thanks for the contribution!

PR #28, opened 2016-04-18
Fix apache issue causing build crashes on restart

dsifford:
Apache is fussy about starting if a pid already exists in the `/var/run/apache2` directory. This change removes that file so that it won't crash with the following error message:

```sh
httpd (pid1) already running
```


karellm:
Thanks 👍

karellm:
Just deployed a new image `0.5.0`.

PR #31, opened 2016-05-16
Add more debug options

dsifford:
Hello again,

I've found these two debug options to be handy while working on some personal projects. Thought you might think so too.

More on them [here](https://codex.wordpress.org/Debugging_in_WordPress#WP_DEBUG_LOG) (if you weren't already aware of them).

Cheers 🍻

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,9 @@ RUN curl \
         -o /run.sh https://raw.githubusercontent.com/visiblevc/wordpress-starter/master/run.sh \
     && chmod +x /usr/local/bin/wp /run.sh \
     && sed -i "s/AllowOverride None/AllowOverride All/g" /etc/apache2/apache2.conf \
-    && a2enmod rewrite
+    && a2enmod rewrite \
+    && service apache2 restart \
+    && wp cli update --nightly --yes --allow-root
 
 # Run the server
 EXPOSE 80 443
--- a/run.sh
+++ b/run.sh
@@ -4,6 +4,8 @@
 [ "$DB_PASS" ]  || DB_PASS='root'
 [ "$THEMES" ]   || THEMES='twentysixteen'
 [ "$WP_DEBUG" ] || WP_DEBUG='false'
+[ "$WP_DEBUG_LOG" ] || WP_DEBUG_LOG='false'
+[ "$WP_DEBUG_DISPLAY" ] || WP_DEBUG_DISPLAY='false'
 [ "$ADMIN_EMAIL" ] || ADMIN_EMAIL="admin@${DB_NAME}.com"
 [ "$SEARCH_REPLACE" ] && \
   BEFORE_URL=$(echo "$SEARCH_REPLACE" | cut -d ',' -f 1) && \
@@ -29,6 +31,8 @@ core config:
   dbhost: db:3306
   extra-php: |
     define('WP_DEBUG', ${WP_DEBUG,,});
+    define('WP_DEBUG_LOG', ${WP_DEBUG_LOG,,})
+    define('WP_DEBUG_DISPLAY', ${WP_DEBUG_DISPLAY,,})
 
 core install:
   url: $([ "$AFTER_URL" ] && echo "$AFTER_URL" || echo localhost:8080)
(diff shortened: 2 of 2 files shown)


dsifford:
Testing now to make sure that having `WP_DEBUG_DISPLAY` default to true doesn't cause any errors.

Otherwise, if you set `WP_DEBUG` to true, you'd have to also explicitly set `WP_DEBUG_DISPLAY` to true as well to match the wordpress defaults (which would get annoying)

dsifford:
After testing for wayyy too long, only to realize that I forgot semicolons on the new settings I added (doh!), I realized two other things that could probably be added to make this better.

1. ) add service apache2 restart to the dockerfile (otherwise you'll get barked at in the logs)
2. ) update to wp-cli nightly in the dockerfile (which has a lot of bugfixes that the current version doesn't have).

Fixing the semicolons now and adding the two other things I mentioned. Feel free to reject if you don't want it.

karellm:
@dsifford I will review this shortly

karellm:
@dsifford Sorry it tooks so long, kinda feel between the cracks!

dsifford:
@karellm Hey, no problem at all.

I've actually been meaning to reach back out to you. I made a from-scratch nginx production build that I'm now using on 2 sites that I was hoping you might want to critique/collab on.

The build has felt pretty strong over the last month or so. The sites are palpably faster (pagespeed > 90/100 on both sites). Here's a quick rundown of the details.
- NGINX server (leveraging [easyengine.io](www.easyengine.io) -- They've spent like 5 years perfecting a wordpress setup.. No sense in reinventing the wheel)
- PHP 7.0
- fastcgi cache
- Adminer drop-in on port 22222 (phpmyadmin replacement)

https://github.com/dsifford/wordpress/tree/nginx

Let me know what you think 👍

dsifford:
Proviso: The initial build takes like 10-15 minutes. EasyEngine runs completely in series using Python.... So, yeah.. Slow.

After the initial build, startup takes 10 seconds. 

I still use the apache version for quick one-off tweaks / plugin development. The NGINX one I use exclusively for production.

karellm:
@dsifford It looks great. I've actually been thinking of switching from Apache to nginx. If you'd like to tackle that in this repo so dev and prod are closer, I'm all for it!

I would also love to see some documentation of your deploy process. I've been meaning to deploy our site to production using Docker but failed to find the time. You seem to have quite some experience on this and would love to pick your brain!

dsifford:
@karellm If you were seriously considering switching over to nginx, then this repo would have to mirror the nginx build I created. The apache setup and nginx setup are so fundamentally different that none of the setup from the current build here is translatable to nginx. 

One thing that I have still not yet figured out with our nginx production builds is obtaining SSL certificates through a proxy server on the cloud host (we use DigitalOcean). Otherwise, the build is shippable as-is.

Re: Deployment -- I wrote a somewhat site-agnostic automatic deploy script for DigitalOcean that some of our less tech-savvy folks have been able to pick up and use. It's a pretty neat setup actually. Here's a quick overview: (this assumes that you have a `.env` file in your root directory with secrets and access tokens inside it).

1. The only input that it needs is your site name, your site name + TLD, and an SSH password.
2. It builds to digitalocean (region NYC1 so that block storage volumes are available) on ubuntu xenial using docker machine.
3. It automatically associates the domain name with the droplet after it's created + sets the A records.
4. After the droplet is created, it creates you a login user with sudo permissions.
5. It pulls your production files into the droplet using rsync and stores them in a data volume.
6. Finally, it builds the site using Docker Compose.

Probably the best part is that it allows other Docker Machine users to attach to the droplet after it's been created using the Docker generic driver (something that I've found to be difficult without some type of automation).

There's probably a lot more stuff that I'm leaving out. You can check it out [here](https://github.com/AcademicLifeInEM/ALiEMU/blob/master/lib/scripts/server-deploy)

I still haven't documented any of it. I was planning on taking it and making it completely agnostic to our project so that others could use it, but I haven't had time to do it yet.

One little proviso: The build script has two dependencies: [jq](https://stedolan.github.io/jq/) and [sshpass](https://gist.github.com/arunoda/7790979)

dsifford:
Now that I think about it, I suppose it would be possible to make an nginx build similar to the current apache build if we leave out all the production optimizations (namely: fastcgi caching) I have built into the production nginx build that I made.

That's probably a good idea.

PR #39, opened 2016-09-19
Detach from docker container on start.

heiso:


karellm:
@heiso Can you at least explain why you think it is better? Wordpress is used by lots of low tech people and detaching the container will mean they loose the output and need another commmand to see it. Not sure this is an improvement.

heiso:
Oh yeah, my PR isn't so documented sorry ^^'
Yeah, that's right, I can understand your point of view. Do what you want with this PR, I thought it was a mistake.

karellm:
I think that people that know docker will have no problem detaching their container so I will close it. Thanks!

PR #40, opened 2016-09-21
Various changes

dsifford:
I was meaning to update this for a little while. Finally got around to it. Feel free to tear this apart.

### Changes:
- Broke apart run.sh into separate functions.
- Add new URL plugin format which allows for better checking after
initial build.
- Add `VERBOSE` environment variable option which allows the build to
run verbosely if one chooses.
- Prettier logging.
- Add `--activate` after plugin installs so that all plugins start out
activated.
- Suppress warning for cases where /data directory isn't volumed
- Correct for missing trailing commas in PLUGINS and THEMES
- Activate, but don't install, locally volumed plugins

### BREAKING CHANGE

You are now required to explicitly mention all plugins and themes you depend on in the compose file, **including locally volumed plugins/themes!**

The format for locally volumed plugins/themes is `[local]my-plugin-slug` and `[local]my-theme-slug` (slug is the directory name by convention). 

This is required so that the build can still be aware that you need those when it goes in to look for orphaned plugins/themes. 

Without adding this to your docker-compose file, the build will think your volumed files are orphaned and it will delete them.... Full stop.... Repeat... _will delete them_.

### Example:
```yml
version: '2'
services:
    wordpress:
        image: dsifford/wordpress:apache-next
        volumes:
            # Database
            - ./data:/data
            # Plugins
            - ./wp-content/plugins/my-plugin-one:/app/wp-content/plugins/my-plugin-one
            - ./wp-content/plugins/my-plugin-two:/app/wp-content/plugins/my-plugin-two
            # Themes
            - ./wp-content/themes/my-theme:/app/wp-content/themes/my-theme
        environment:
            DB_NAME: wordpress
            DB_PASS: root
            DB_USER: root
            PLUGINS: >-
                akismet,
                [local]my-plugin-one,
                [local]my-plugin-two,
            THEMES: >-
                [local]my-theme,
    db:
        image: mariadb:10
        expose:
            - "3306"
        environment:
            MYSQL_ROOT_PASSWORD: root
        volumes:
            - data:/var/lib/mysql
```


docker-compose.yml
```diff
@@ -12,9 +12,11 @@ services:
     environment:
       DB_NAME: wordpress
       DB_PASS: root
-      PLUGINS: akismet
+      PLUGINS: >-
+        academic-bloggers-toolkit,
```
karellm: I'm not a fan of having default plugins unless there is a reason?

run.sh
```diff
@@ -13,15 +20,206 @@ WP_DEBUG=${WP_DEBUG:-'false'}
...
+  h3 "Installing needed themes"
+  WP theme install --quiet "${THEME_LIST[@]}"
+  STATUS $?
 }
 
-# Configure wp-cli
-# ----------------
+# Config Utility Functions
+# -------------------------
+generate_config_for() {
```
karellm: I think I would prefer two functions here. I tend to like SRP.

run.sh
```diff
@@ -13,15 +20,206 @@ WP_DEBUG=${WP_DEBUG:-'false'}
...
+    WP core multisite-convert |& loglevel
+    STATUS ${PIPESTATUS[0]}
+  else
+    h3 "Multisite not found. Skipping..."
+    STATUS SKIP
+  fi
+
+  # Operations to perform on first build
+  # ------------------------------------
+  if [ -d /app/wp-content/plugins/akismet ]; then
```
karellm: This is not reliable, some install need `akismet`. We could `touch .first_install` or similar

karellm: Also I would run that prior `check_plugins` so if the user actually want `akismet` we don't delete it.

run.sh
```diff
@@ -13,15 +20,206 @@ WP_DEBUG=${WP_DEBUG:-'false'}
   AFTER_URL=$(echo "$SEARCH_REPLACE" | cut -d ',' -f 2) || \
   SEARCH_REPLACE=false
 
-ERROR () {
-  echo -e "\n=> $(tput -T xterm setaf 1)$(tput -T xterm bold)ERROR$(tput -T xterm sgr 0) (Line $1): $2.";
-  exit 1;
+
+main() {
+
+  generate_config_for wp-cli
```
karellm: I don't think this warrant a function. The function as a `if` statement and does 2 things. I prefer inlining the code rather than having wrapper function with obscure api and logic.

run.sh
```diff
@@ -13,15 +20,206 @@ WP_DEBUG=${WP_DEBUG:-'false'}
...
+  # --------------
+  h2 "Waiting for MySQL to initialize..."
+  while ! mysqladmin ping --host=db --password=$DB_PASS --silent; do
+    sleep 1
+  done
+
+  h1 "Begin WordPress Configuration"
+
+  h3 "Generating wp.config.php file"
+  generate_config_for wordpress
```
karellm: Same as above

docker-compose.yml
dsifford: I just added those there as an example of the two formats that are available.

run.sh
dsifford: Agreed w/ both of your concerns.

run.sh
dsifford: No problem.. will revert later on

run.sh
dsifford: ditto above

dsifford:
@karellm 

Check out these changes when you get time. I went a bit nuts with refactoring, but I think it generally is a more logical progression though the build process.

Most notably, I removed the need for `touch .first_build` checks. It works appropriately no matter if it's the first build or a rebuild.

Still need to test this a bit more, but the few tests that I have done seem to work well. 👍

gulpfile.js
```diff
@@ -18,7 +18,7 @@ var awspublish   = require('gulp-awspublish');
 var autoprefixer = require('gulp-autoprefixer');
 
 // Configuration
-var config = JSON.parse(fs.readFileSync('./gulp.json'));
+var config = require('./package.json').gulp
```
dsifford: @karellm This was bugging me... Feel free to revert if you hate it.

gulpfile.js
```diff
@@ -18,7 +18,7 @@ var awspublish   = require('gulp-awspublish');
 var autoprefixer = require('gulp-autoprefixer');
 
 // Configuration
-var config = JSON.parse(fs.readFileSync('./gulp.json'));
+var config = require('./package.json').gulp;
```
dsifford: Changed to this format for performance / best practice purposes.

README.md
```diff
@@ -97,44 +98,49 @@ volumes:
...
 ### Working with databases
 
 If you have an exported `.sql` file from an existing website, drop the file into the `data/` folder. The first time you run the container, it will detect the SQL dump and use it as a database. If it doesn't find one, it will create a fresh database.
 
 If the SQL dump changes for some reason, you can reload the database by running:
 
-```
-docker-compose run wordpress wp db import "$(find /data/*.sql | head -n 1)" --allow-root
+```sh
+docker exec wordpress /bin/bash "wp db import $(find /data/*.sql | head -n 1) --allow-root"
```
dsifford: See #41

README.md
```diff
@@ -97,44 +98,49 @@ volumes:
 - `WP_DEBUG` (optional): Defaults to `false`
 - `WP_DEBUG_DISPLAY` (optional): Defaults to `false`
 - `WP_DEBUG_LOG` (optional): Defaults to `false`
-- `THEMES` (optional): Comma-separated list of themes you want to install.
-- `PLUGINS` (optional): Comma-separated list of plugins you want to install.
+- `THEMES` (optional): Comma-separated list of themes you want to install in either of the following forms:
+  - `theme-slug`: Used when installing theme direct from WordPress.org.
+  - `[theme-slug]http://themesite.com/theme.zip`: Used when installing theme from URL.
```
dsifford: Added split format for themes as well so people can install from URLs

docker-compose.yml
karellm: Do you feel strongly about that? I think it is more the role of the documentation. If you follow the getting started steps, you will end up with plugins you didn't ask for. I personally don't think it is a good thing since people might wonder if it is a requirement to the project.

docker-compose.yml
```diff
@@ -1,6 +1,9 @@
 version: '2'
 services:
   wordpress:
+    # build:        # Disregard this - used for project development purposes
```
karellm: I feel pretty strongly against commented out code. Can we avoid that?

run.sh
```diff
@@ -45,170 +45,338 @@ core install:
...
+  h3 "Generating wp.config.php file"
+  rm -f /app/wp-config.php
+  WP core config |& loglevel
+  STATUS "${PIPESTATUS[0]}"
+
+  h2 "Checking database"
+  check_database
+
+  # Make multisite
+  # NOTE: This will likely cause issues down the road.
```
karellm: Do you have something in mind? I have used multisites before and they don't defer that much from regular installs.

dsifford: I have precisely zero experience with multisite so I'm open to any suggestions you might have

docker-compose.yml
dsifford: Yep -- I'll cut that out just before the merge 👍

dsifford:
AFAIK, all the issues that you've mentioned should be taken care of now. 

Let me know if I've missed anything.

dsifford:
@karellm Any chance I can get you to change your mind on automatic builds? Reason being: It would be nice to be able to tell people who are posting issues to try building a PR branch to see if those changes resolve their problems.

karellm:
Yes let's do it. Can you point me to a good resource to set it up?

dsifford:
https://docs.docker.com/docker-hub/builds/ 👍

dsifford:
@karellm This is caught up with the recent update to `master`. What did you want to do with this?

Also, were you ever able to get auto builds set up?

dsifford:
@karellm can you add me as a contributor to the dockerhub page? If so, I'll set up automated builds.

dsifford:
Another benefit of automated builds...

Search term = "wordpress"; 4th result

image

dsifford:
Hate to be a pest about this, but I'm sort of in need of this today to get some of my work done that I've been putting off. If we can't merge today, I'm going to have to copy this over to my fork and build that way

dsifford:
Scratch my last... I copied this to my fork and I'm good for now.

Also, I found a small bug that I need to fix before merging -- it'll introduce a potentially breaking API change, but I'm not sure how else to go about it... More on this later

run.sh
```diff
@@ -60,7 +60,7 @@ main() {
   # Wait for MySQL
   # --------------
   h2 "Waiting for MySQL to initialize..."
-  while ! mysqladmin ping --host=$DB_HOST --password="$DB_PASS" --silent; do
+  while ! mysqladmin ping --host="$DB_HOST" --password="$DB_PASS" --silent; do
```
dsifford: This needed to be quoted to avoid shell expansion

run.sh
```diff
@@ -134,7 +134,7 @@ check_database() {
     STATUS "${PIPESTATUS[0]}"
 
     # If an SQL file exists in /data => load it
-    if [ "$(stat -t /data/*.sql >/dev/null 2>&1)" ]; then
+    if [[ "$(find /data -name '*.sql' | wc -l)" != "0" ]]; then
```
dsifford: `find` is more reliable in this situation

run.sh
```diff
@@ -197,13 +197,27 @@ check_themes() {
       h3 "($i/$theme_count) '$theme_name' not found. Installing"
       WP theme install --quiet "$theme_name"
       STATUS $?
+      ((i++))
```
dsifford: forgot to increment the counter in circumstances where the loop ends early

run.sh
```diff
@@ -197,13 +197,27 @@ check_themes() {
...
+      continue
+    fi
+
+    # Locally volumed themes
+    if [[ $theme_name =~ ^\[local\] ]]; then
```
dsifford: ## BREAKING CHANGE

You are now required to explicitly mention all plugins and themes you depend on in the compose file, **including locally volumed plugins/themes!**

The format for locally volumed plugins/themes is `[local]my-plugin-slug` and `[local]my-theme-slug` (slug is the directory name by convention). 

This is required so that the build can still be aware that you need those when it goes in to look for orphaned plugins/themes. 

Without adding this to your docker-compose file, the build will think your volumed files are orphaned and it will delete them.... Full stop.... Repeat... _will delete them_.

@karellm if you have any better ideas as to how to go about this, I'm all ears. I can't think of a better way for this.

run.sh
```diff
@@ -197,13 +197,27 @@ check_themes() {
...
+      ((i++))
       continue
     fi
 
     # If $theme_name matches a URL using the new format => set $theme_name & $theme_url
+    # If url is from github, correct for the branch name in the plugin name
     if [[ $theme_name =~ ^\[.+\]https?://[www]?.+ ]]; then
       theme_url=${theme_name##\[*\]}
       theme_name="$(echo "$theme_name" | grep -oP '\[\K(.+)(?=\])')"
+      if [[ $theme_url =~ github.com ]]; then
```
dsifford: The URL for plugins installed from github.com is appended with the branch name. This corrects for that so that the build can still check if those plugins are installed or not.

In other words: If I'm installing `akismet` from the master branch of the akismet repo, the url will look like this: `akismet-master`. This fixes it to look like this: `akismet`.

run.sh
```diff
@@ -275,13 +289,27 @@ check_plugins() {
...
       WP plugin install --activate --quiet "$plugin_name"
       STATUS $?
+      ((i++))
+      continue
+    fi
+
+    # Locally volumed plugins
+    if [[ $plugin_name =~ ^\[local\] ]]; then
+      plugins["${plugin_name##*]}"]="${plugin_name##*]}"
+      h3 "($i/$plugin_count) '${plugin_name##*]}' listed as a local volume. SKIPPING..."
```
dsifford: ## BREAKING CHANGE

(see above comments)

run.sh
```diff
@@ -296,8 +324,8 @@ check_plugins() {
       STATUS $?
       # Pretty much guarenteed to need/want 'restful' if you are using 'rest-api'
       if [ "$plugin_name" == 'rest-api' ]; then
-        h3 "       Installing 'restful' WP-CLI package"
-        WP package install danielbachhuber/resftul --quiet
+        h3 "($i.5/$plugin_count) Installing 'restful' WP-CLI package"
+        wp package install wp-cli/restful --quiet --allow-root
```
dsifford: repo was moved

run.sh
dsifford: perhaps `volume` might be more intuitive than `local`? Dunno...

dsifford:
@karellm You can currently test this branch by pulling `dsifford/wordpress:apache-next`

dsifford:
Added 3 fixes just now for some other small bugs that I ran into

dsifford:
@karellm Pushed a small update just now + updated original PR message with example.

dsifford:
@karellm just confirming: You're okay with the breaking change as-is?

karellm:
@dsifford Yes I am. I'm actually making sense of all the changes right now and see how it ports back to our setup.

Quick comment for the future, please don't merge master into PRs but use rebase. It makes cleaning up the git history so much easier when comes the time to merge.

karellm:
Setting `VERBOSE` to `'false'` will cause `check_database` to always pass. This caused me quite some headaches when testing the new setup!

I squashed your commits and added one for suggested changes. Mostly, this standardize the output a step further. The `h1` felt weird in the middle of `h2` and `h3`. Unless you feel strongly about I prefer without it. Let me know what you think.

dsifford:
Whoops, sorry about the merge commit. Didn't think about that!

I'm cool with your suggestions for the most part. I'll add some comments inline in just a sec to two specific things that I noticed, but other than that I'm good.

> Setting VERBOSE to 'false' will cause check_database to always pass.

Explicitly setting it to false in the compose file or implicitly in the run script? I haven't had that issue.

> The h1 felt weird in the middle of h2 and h3. Unless you feel strongly about I prefer without it.

I think we should keep it at least at the end to provide some type of delineation between builds (screenshot below shows logs for an initial build, a stop, and a rebuild). My feelings aren't totally strong on this, but that'd be my vote if it came down to one. I do agree having one stray `h1` in the middle looks out of place though.


image

run.sh
```diff
@@ -51,24 +51,25 @@ main() {
   # Download WordPress
   # ------------------
   if [ ! -f /app/wp-settings.php ]; then
-    h3 "Downloading wordpress"
+    h2 "Downloading wordpress"
     chown -R www-data:www-data /app /var/www/html
     WP core download |& loglevel
+    h3 "Wordpress downloaded"
```
dsifford: I'd reword these two to something along the lines of `Initializing..` and `Downloading WordPress` simply because the `STATUS "${PIPESTATUS[0]}"` can potentially show a pass or a fail. "WordPress downloaded" to me denotes it will always pass.

run.sh
```diff
@@ -51,24 +51,25 @@ main() {
...
     sleep 1
   done
 
-  h1 "Begin WordPress Configuration"
-
-  h3 "Generating wp.config.php file"
+  h2 "Generating wp-config.php file"
   rm -f /app/wp-config.php
   WP core config |& loglevel
+  h3 "wp-config.php created"
```
dsifford: I'd move the text from the above `h2` to this `h3` block and remove the `h2` block for the same reason as above. This has the potential to pass or fail and the wording here to me sounds like it's always passing.

karellm:
@dsifford I just amended some changes to the commit.

If you set `VERBOSE: 'false'` in your docker-compose ENV, it works ok for you?

dsifford:
Ah, I see.. yeah that's probably why you had issues. 

The `VERBOSE` flag is supposed to only be set to `'true'` if you need it. Otherwise, leave it out. Setting it to `'false'` in compose sets the actual implementation of `VERBOSE` in `run.sh` to a string, which will always evaluate to true.

dsifford:
I suppose I can fix that if you'd like

karellm:
@dsifford I think it would be better fixed yes. Thanks!

dsifford:
Will do... Busy with other things atm, but I should be able to get to it in the next couple hours

dsifford:
@karellm I lied. Now that I'm looking at this, I'm not sure what the error is you were describing Re: `VERBOSE='false'`. I can't get it to fail any way I try. Can you explain what's happening a bit more?

karellm:
@dsifford I couldn't reproduce either. My bad! I amended couple things. I invite you to test this branch and if you are happy with it I will merge that today.

PS: Try creating a new issue. I added a `ISSUE_TEMPLATE.md` file at the root of the project so people are a little more thorough when filing issues! Feel free to add things to it if I forgot anything.

dsifford:
@karellm Looks good to me. I made 2 changes to the capitalization of "WordPress". Other than that, looks good.

I've been using this branch now for over a month so I can comfortably say the merge is ready 👍 

Re: `ISSUE_TEMPLATE.md` -- Looks pretty good. One thing I've been burned on is that people still don't care to read the directions and they leave all the stock data there and it clutters everything.

One thing I'd suggest doing is placing the all instructions for the issue submitter in html comment blocks so that it isn't rendered with the issue when submitted.

So, in other words:

This...
```html
These are the instructions.

Blah blah blah
```

Should be this...
```html
<!--
These are the instructions.

Blah blah blah
-->
```

dsifford:
Also -- I made no changes to the changelog. Just an FYI.

karellm:
@dsifford this is deployed as 0.10.0. 

For the comments, since it adds works on the issue publisher to remove them, I would wait a little and see if people actually post without changing the content. Then I'm all for adding the comments. In the meantime, I've made it clearer then need to change the content.

PR #57, opened 2016-12-05
Add support for php7.0 and php5.6

dsifford:
This PR adds support for both `php7.0` and `php5.6`.

To do this, two new Dockerfiles had to be created. The base image for both of these images is from the official PHP image from the Docker Registry (which uses debian as its base OS, rather than ubuntu).

I changed the directory structure to better contain these new files.

We'll have to now maintain 4 tags for this, which I propose should be...
- `latest` and `php7.0` for the PHP 7 image.
- `php5.6` for the PHP 5.6 image.
- `legacy` for the PHP 5.5.9 image.

The tag `legacy` should be used because the base image for that Dockerfile hasn't been updated in over a year. It's based from a tutumcloud image that appears to be abandoned. We should not be encouraging the use of that image.

Anyway, I'll leave it at that for now. Please critique my changes and let me know if you have any questions.

(Note: I didn't address the issue that we currently have open in this PR. I did, however, add the `[local]` prefix to the documentation for the time being).


docker-compose.yml
```diff
@@ -1,21 +1,17 @@
...
     volumes:
-      - ./data:/data
+      - data:/data
     environment:
       DB_NAME: wordpress
       DB_PASS: root
   db:
-    image: mysql:5.7 # or mariadb:10
-    ports:
-      - 3306:3306
```
dsifford: Two things here:
1. `PORTS` was removed because apparently in compose `v2` all services can access each other by default. So this is just redundant.
2. Switched to recommending `mariadb` because everybody seems to say that's the faster of the two, and also because I'm not a fan of oracle.

php5.6/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM php:5.6-apache
+MAINTAINER Derek P Sifford <dereksifford@gmail.com>
+
+# Install base requirements & sensible defaults + required PHP extensions
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        less \
+        libpng12-dev \
+        libjpeg-dev \
+        mariadb-client \
```
karellm: We might want mysql client here too no?

php7.0/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM php:7.0-apache
+MAINTAINER Derek P Sifford <dereksifford@gmail.com>
+
+# Install base requirements & sensible defaults + required PHP extensions
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        less \
+        libpng12-dev \
+        libjpeg-dev \
+        mariadb-client \
```
karellm: Same

php5.6/Dockerfile
dsifford: They're cross-compatible. `mariadb-client` is 100% compatible with MySQL and `mysql-client` is 100% compatible with MariaDB. Only one is necessary.

karellm:
@dsifford Before we merge I want to address the tagging system. I think that we need to combine versions and php version. Something like:

`latest`, `latest-php7.0`, `0.12.0`, `0.12.0-php7.0`
`latest-php5.6`, `0.12.0-php5.6`

Finally I'm fine removing the legacy image.

dsifford:
@karellm Ah, yeah I was meaning to ask what you think the best way to go about semver would be for maintaining multiple Dockerfiles.

Your suggestion seems good to me.

karellm:
@dsifford Deployed. I updated the instructions for deploying the images: 

https://github.com/visiblevc/wordpress-starter/wiki/Development

dsifford:
@karellm Nice.

I never checked, did you say I have access to the dockerhub project?

karellm:
@dsifford You should yes

dsifford:
Cool. I'll check it out a bit later

PR #58, opened 2016-12-05
Convert environment variables in readme to tabular format

dsifford:


karellm:
Looks a lot cleaner, thanks!

PR #59, opened 2016-12-06
Critical Fix

dsifford:
Tried using the build this evening and it didn't work properly (even after working all day yesterday? Go figure). Did some digging and it looks like the root of the problem is that the zip extension isn't enabled so wp-cli is unable to unzip the initial wordpress download.

Fixed that.

Please test on your end to make sure this works for you.

### DIFF
--- a/php5.6/Dockerfile
+++ b/php5.6/Dockerfile
@@ -14,7 +14,7 @@ RUN apt-get update \
         zip \
     && rm -rf /var/lib/apt/lists/* \
     && docker-php-ext-configure gd --with-png-dir=/usr --with-jpeg-dir=/usr \
-    && docker-php-ext-install gd mysqli opcache
+    && docker-php-ext-install gd mysqli opcache zip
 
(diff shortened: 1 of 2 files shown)


dsifford:
@karellm just tried again and this is still broken. Might need to roll back until I can figure out why it's not working once uploaded to dockerhub.

dsifford:
Also, I have no access to the dockerhub account. You did deploy this latest change, correct?

dsifford:
I know you said earlier you were concerned about github repo access with automated builds. Would you be willing to move this project somewhere else so that we can enable them? It makes life so much easier.

karellm:
@dsifford I forgot to push it, you can try `0.12.1` now.

dsifford:
@karellm What the heck is going on? I just tried `latest-php7.0` and say's it's running php 5.6...

image

I have no idea what's going on. I give up for now.

karellm:
@dsifford I think it is my mistake. I retags the images, should be good now.

dsifford:
@karellm You're trying to give me a heart attack, aren't you 😂 

image

All good 🙌

PR #70, opened 2017-01-16
add certbot to both images for SSL cert automation

dsifford:
This PR simply adds certbot to both images so that users who choose to run the image in production can obtain and renew certificates easily.

To obtain new certificates, open an interactive session in the wordpress service, run `certbot --apache`, and then follow the prompts.

To renew certificates, open an interactive session in the wordpress service and run `certbot renew`.

Alternatively, you can set the renew task up as a cronjob which runs every 90 days.

There's not really a great way to add it into the build script because it requires user interaction using interactive terminal prompts.

Let me know what you think @karellm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 This changelog is for the Docker image. There might be changes to the gulpfile that are not listed here.
 
-## 0.12.2 - latest
+## 0.13.0 - latest
+- Add `certbot` into the image to allow for easily obtaining and renewing SSL certificates
+
+## 0.12.2
 - Add missing `exif` php extension to handle WordPress media upload metadata parsing.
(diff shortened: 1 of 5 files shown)


karellm:
This looks good to me! You can merge as you see fit :)

dsifford:
Curious if you think it would be worth it to offload the cert automation to a separate service vs keeping it within the wordpress service?

Trade off would be separation of concerns vs ease of use. (Certbot requires python and a bunch of libraries to operate)

karellm:
Both are fine imo, I really don't have a strong preference here.

dsifford:
@karellm Cool. We'll keep as-is for now until someone complains.

Just updated the readme and changelog. Bumped to version 0.13.0. Gonna merge this into master now and push to Dockerhub. 👍

PR #76, opened 2017-01-19
Add composer to the container, fix #74

### LINKED ISSUE #74: Suggestion: add Composer to the image
Having PHP's Composer pre-installed in the image and adding a 'REQUIRE` property in the `docker-compose.yml` would be a very nice addition to this project I believe.

I'd like to use wordpress-starter to unit-test a plugin of mine. Having Composer pre-installed would allow me to quickly install PHPUnit and test my plugin.

I might give a PR a try, but I'd prefer to be sure that it has a chance of being accepted.

Thank you in advance :-)

louisremi:
Hey guys,
Here's my attempt to add Composer to this container.
This is my first attempt at writing bash scripts, I've only started using Docker two weeks ago, and I've never used Composer before either, so please bear with me :-)

Please note: I've only updated the Docker file in `php7.0/`, and it accesses the `run.sh` that's hosted on my fork of the project. I'll fix that once you've tested building and running this image locally :-)

I successfully installed and loaded phpunit using composer. I'll continue testing it during the day.


louisremi:
(I finally managed to add `/app/vendor/bin` to the container's path, so I'm now able to use this project to unit-test my plugins \o/)

run.sh
```diff
@@ -346,6 +349,23 @@ check_plugins() {
...
 
+check_packages() {
+  if [[ ! "${REQUIRE-}" ]]; then
+    h3 "No package dependencies listed"
+    STATUS SKIP
+    return
+  fi
+
+  h3 "Installing composer packages"
+  touch /app/composer.json
```
dsifford: What's the benefit of this opposed to just voluming a `composer.json` file in the app directory?

Seems like these next few lines could be quite fragile.

run.sh
```diff
@@ -346,6 +349,23 @@ check_plugins() {
...
+  echo "{\"require\":{" > /app/composer.json
+  echo "$REQUIRE" |tr '\n' '\r' |sed -r 's/(^\s*|,?\s*$)/"/g' |sed -r 's/\s*(:|,)\s*/"\1"/g' |tr ',' ',\n' >> /app/composer.json
+  echo "}}" >> /app/composer.json
+
+  cd /app || exit
```
dsifford: `WORKDIR` is already set to `/app` so this shouldn't be needed

run.sh
```diff
@@ -346,6 +349,23 @@ check_plugins() {
...
+  composer install
+  rm -f /app/composer.json
```
dsifford: Is this necessary?

php7.0/Dockerfile
```diff
@@ -49,6 +51,9 @@ RUN curl \
 RUN mkdir -p /app && rm -fr /var/www/html && ln -s /app /var/www/html
 WORKDIR /app
 
+# Add Composer's vendor/bin/ to the PATH
+ENV PATH "$PATH:/app/vendor/bin"
```
dsifford: This creates another layer in the image. Can we move this to the run script?

run.sh
louisremi: Voluming a composer.json is still possible, but with this feature, you need one less config file, and all the dependencies of your wordpress are listed in a single place.

run.sh
louisremi: Alright

php7.0/Dockerfile
louisremi: I've spent quite some time trying to modify the PATH from `run.sh` but all of my attempts failed. It seems to be only feasible from the Dockerfile, but feel free have a look at my different PATH related commits, and try a different solution.

run.sh
louisremi: Also, composer.json could be volumed anywhere: in `/app`, `/app/wp-content`, `/app/wp-content/plugins`, `/app/wp-content/plugins/any-plugin`. So detecting this file to automatically `composer install` it is quite a challenge. Detecting REQUIRE is far easier.

run.sh
louisremi: No, I'll remove it.

run.sh
dsifford: No strong feelings on it. I just think it would be easier from a maintenance perspective.

Additionally, going this way, you wouldn't need to set `REQUIRE`. Just have the run script check for the existence of `composer.json`.

I'll let @karellm be the judge on which way to go for that.

php7.0/Dockerfile
dsifford: What part of `run.sh` requires `/app/vendor/bin` to be set in the path?

run.sh
louisremi: Please see my second comment above: detecting REQUIRE is far easier than finding composer.json

php7.0/Dockerfile
louisremi: run.sh does not require this. This is only required to use modules installed by Composer from the command line.
Without this, I can install the phpunit package and write unit test files, but I cannot run those test using `phpunit` from the command line :-/

dsifford: Did you use `export PATH="$PATH:/app/vendor/bin` in `run.sh`? `export` being the important part.

run.sh
dsifford: I think that if someone is advanced enough to know how to use composer, they should be able to follow a simple instruction to volume the `composer.json` into the root of the `/app` directory. 

We're already at a point of having way too many environment variable options. Makes things chaotic to keep adding more and more IMHO.

Deferring the judgement call for this to @karellm.

php7.0/Dockerfile
louisremi: Yes, see https://github.com/visiblevc/wordpress-starter/pull/76/commits/1abeea08d87ccee30df969f113040aeaf966e726

run.sh
karellm: I agree we have too many ENV variables and that this one might confuse some people. My tendency would also have been to mount the file at a specified location (I wouldn't do a lookup for the file).

That said, I also see the argument of consistency and having everything configured in one place.

What I really hate though is generating the `compose.json` file from `run.sh`. Isn't there a `install` command for composer like for npm that would do that for us??

run.sh
```diff
@@ -346,6 +349,25 @@ check_plugins() {
...
 
+check_packages() {
+  # If a composer.json file exists in /app => install it
+  if [[ -f "/app/composer.json" ]]; then
+    h3 "Installing packages listed in composer.json"
+    composer install
+    return
+  fi
+
+  if [[ "${REQUIRE-}" ]]; then
```
karellm: @louisremi I think this should be exclusive: Either install form the file or from the REQUIRE but not both.

louisremi: This doesn't install from both, but gives both options to developers: they can start by adding their packages to their docker-compose.yml, and later move it to a separate composer.json file when they feel this clutters their ENV too much. Using REQUIRE will be easier for beginners (me, for example), while using a separate composer.json will be easier for people used to Composer.

karellm: Yes but let's say you have a `composer.json` file and left a REQUIRE by mistake. It will overwrite your `composer.json` file. I don't think that's a behaviour we want.

louisremi: Installing volumed composer.json and then adding packages from REQUIRE is totally feasible as well.

louisremi:
@karelim what proposed change did you prefer?

louisremi:
With these last two commits and after a week of testing, I feel this PR is no longer a WIP :-)

karellm:
@dsifford How do you feel about this in the end?

dsifford:
@karellm I'll personally never use it. So for my own personal use-case, it'll just be extra bloat.

I'll surrender my opinion to what the majority decides however.

Spone:
I'm also using Composer to install some dependencies for my theme, so I'm all for merging!

:+1: @louisremi

dsifford:
This got a bit outdated from master and I think this isn't something that fits the specific goals of this project. So I'm gonna close this and suggest maybe forking the project and adjusting to your needs.

PR #78, opened 2017-01-21
add SERVER_NAME variable

dsifford:
Hey there, hypocrite here! 

I know I just got done saying we're env variable overloaded (I think yesterday?), but I deployed a site in production last night and the single-biggest annoyance was that I had to go into the container, navigate to the apache config file, and set the ServerName directive to the domain name prior to running `certbot --apache` (otherwise, things would get screwy with certbot).

This PR allows for users to set the `ServerName` directive in the `000-default.conf` on build, which negates the need to to anything but `certbot --apache` for getting SSL certificates.

Because of this option, setting `ServerName` to `localhost` was removed from the Dockerfile and placed in the run script.

Let me know what you think @karellm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 This changelog is for the Docker image. There might be changes to the gulpfile that are not listed here.
 
-## 0.13.0 - latest
+## 0.14.0 - latest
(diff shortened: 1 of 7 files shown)


karellm:
@dsifford LGTM 👍  Feel free to deploy!

PR #79, opened 2017-01-24
[WIP] Refactor dockerfile

louisremi:
I wish to refactor our Dockerfiles to make it easier to create alternative images. The current config uses php5.6-apache and php7.0-apache base images, and most code in the Dockerfiles is repeated. Users might want to create derivative images from other base image, for example php7.1-apache

This pull request makes it easy to create such images. I've included examples using php-xdebug base images.

Happy to discuss this PR with you :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,6 @@ wp-content/*
 !wp-content/themes/
 
 # WordPress Starter Dev Files
+php*
 Dockerfile-test
 test.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+MAINTAINER Derek P Sifford <dereksifford@gmail.com>
+
+# Install base requirements & sensible defaults + required PHP extensions
+RUN echo "deb http://ftp.debian.org/debian jessie-backports main" >> /etc/apt/sources.list \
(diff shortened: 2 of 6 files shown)


louisremi:
I wonder if there is a way of not using `--no-cache`, or emptying the cache once before building all images.

dsifford:
@louisremi I'd personally rather just have automated builds on dockerhub. Not discounting the work you did, but it's mostly unnecessary with auto builds. There's an open issue about this now.

I don't mind manually changing the tags in the dockerfiles on updates because that takes about 10 seconds to do.

louisremi:
Yep, I just discovered automated build and was already having the impression this would solve my problem.

louisremi:
So wouldn't it be simpler to have a separate repository to host `Dockerfile` and `run.sh`?
At least for people like me who'd like to setup automated builds on forks of those files..

dsifford:
How would that make it simpler?

louisremi:
You're right, it's not necessary. I posted this after I saw that automated builds tag images according to branch names. But that behavior can be changed to use directories in a single branch. Closing for now. I'll reopen if I find these scripts to be significant time-savers in the future.

PR #82, opened 2017-01-26
Adding xdebug to the image

louisremi:
I've created a new Dockerfile in `php7.0-debug/` that adds xdebug to the image. Unfortunately, despite all my efforts, I never managed to debug any script with it :-(
Any help would be appreciated.


louisremi:
Ok, so I finally managed to get xdebug to work with Atom. I had no idea how xdebug worked, and the fact that it needs to match a local file on your computer is a huge pain in our case, as most of the code is only present in the container: you can't step in a core function of WordPress for instance, as those files are inaccessible by your text editor :-/

dsifford:
@karellm can make the judgement call on this. I don't write php. In fact, I actively do everything I can to avoid writing php.

When dealing with Wordpress, the only purpose of the php I write is to serve JavaScript. 

I'll personally never use xdebug or composer for those reasons.

The heart and soul of this project is to make Wordpress less miserable. This just injects misery IMHO.

louisremi:
I share your distaste for PHP. I was just hoping to find tools similar to those I use in JS. Composer seems okayish, but debugging looks like a huge pain in the ass.

dsifford:
I've found the easiest way to inspect php variables is to either echo them into the page, echo them into a script tag that prints to the js console, or to just pass it to `wp_die()`. 

Agree that it's a gigantic pain.

Similarly, I've found PHPUnit to be equally as horrendous to work with. Feels like working with both hands tied behind your back.

run.sh
```diff
@@ -116,8 +119,12 @@ main() {
 
   h3 "Adjusting file permissions"
   groupadd -f docker && usermod -aG docker www-data
-  find /app -type d -exec chmod 755 {} \;
-  find /app -type f -exec chmod 644 {} \;
+  if [[ "${PHP_ENV}" =~ ^dev ]]; then
+    chmod -R 777 /app
```
karellm: I think that having different permissions in development is a bad idea, it will cause issue when you deploy. whyt @dsifford ?

dsifford: Changing permissions to 777 in any situation is generally always a bad idea. The main issue is ownership, not permissions.

So the real "fix" for this would be to just create a user other than root and bootstrap with that user.

The fix would have to also be followed with a few fixes of the `run.sh` file because right now it's tailored around being root.

dsifford: The canonical user for this would be `www-data`. I can fix this when I get some free time in the next week or so.

karellm:
@dsifford @louisremi What do we do with this? I vote for closing it.

dsifford:
Yeah, we can close. It's a bit too far outdated with HEAD tip of the repo anyway.

PR #85, opened 2017-02-15
support custom db user

hjanuschka:


karellm:
@hjanuschka Why you need this? It is a good habit when contributing to an OS project to give some context and a description to a PR. 

Also any change to the codebase should be reflected in the README. Thanks!

karellm:
@hjanuschka Why you need this? It is a good habit when contributing to an OS project to give some context and a description to a PR. 

Also any change to the codebase should be reflected in the README. Thanks!

hjanuschka:
@karellm alright sorry that was way to fast - going to close the PR - and will re-do it with a few more changes and updated README.

PR #90, opened 2017-04-04
Added DB_HOST explanation

viennaoperator:
Hello,
I was struggling to understand the DB_HOST variable and didn't know that it has to match the service name of the mysql/mariadb container. (because docker-compose automatically links the mysql/mariadb container via name in /etc/hosts within the wordpress container)


karellm:
Thanks @viennaoperator !

PR #98, opened 2017-04-24
Denies Apache from seeing .yml files such as the wp-cli.yml file

joerybruijntjes:
The wp-cli.yml in the application root contains sensitive information needed during setup but never gets removed. Not sure if it needs to exist beyond the initial setup so as a stopgap measure this patch simply disallows Apache to serve any .yml files to the visitor so that information can't leak. If I'm missing some obvious reason why that information would be not fetch-able whilst running in production, let me know!

### DIFF
--- a/php5.6/Dockerfile
+++ b/php5.6/Dockerfile
@@ -44,6 +44,7 @@ RUN curl \
         -o /run.sh https://raw.githubusercontent.com/visiblevc/wordpress-starter/master/run.sh \
     && chmod +x /usr/local/bin/wp /run.sh \
     && sed -i "s/AllowOverride None/AllowOverride All/g" /etc/apache2/apache2.conf \
+    && sed -i 's/<FilesMatch "^\\.ht">/<FilesMatch "^\.ht|\.yml">/g' /etc/apache2/apache2.conf \    
     && a2enmod rewrite expires \
     && service apache2 restart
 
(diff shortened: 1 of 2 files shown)


karellm:
@joerybruijntjes This is a development environment, this isn't intended to be used in production. I don't see the need for this.

dsifford:
@karellm I disagree. This had potential to be disastrous. Kudos on the find @joerybruijntjes, huge help. (I'm using this in production on two sites).

I'm rolling a different fix into the currently open PR I have that targets this in a slightly different way. I'll give you credit for the in the release notes. Thanks again.

@joerybruijntjes If you have any background or familiarity with PHP servers, I welcome any further scrutinization of the codebase. I have basically none.

karellm:
@dsifford If you do use it in production then you can discard my previous comment. I didn't realize people were.

joerybruijntjes:
@karellm Even if nobody was running it in production I would still be in favour of being secure by default, especially if it is such a quick and easy fix. Just look at all the trouble with publicly accessible MongoDB databases that didn't require authentication. One could argue it's on the developer who put such a thing into production without fully grasping the consequences yet that does not make the resulting damage in terabytes of stolen (user)data and such less real. Also, in this case the project readme contains instructions for a production run so I considered someone might try it.

@dsifford Welcome, just glad I wasn't mistaken about it being an issue and that your sites are now patched. Jeej for a slightly safer internet :-) I'm relatively new to Docker but have done my share of server administration - if I find anything else whilst tinkering with the project I'll let you guys know!

PR #101, opened 2017-05-30
Add deploy script for automating the manual deploy process

dsifford:
cc: @karellm


karellm:
Looks great thanks! I always learn new things reading your bash script 👍

PR #106, opened 2017-07-17
Addition to wp-config.php to allow https via reverse proxy

signaleleven:
I tried to put this wordpress container behind a reverse proxy that terminates https requests and reroutes them in clear to a local container (I used traefik, but it could be anything) and the only way to make it work properly (no mixed content) is to allow wordpress to be aware of the fact that the request is a forwarded https request.
I believe that these lines, added to the wp-config.php file, are harmless for other use cases, but crucial for mine. 

Pull if you agree, and thanks for the great project!

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -41,9 +41,9 @@ core config:
     define('WP_DEBUG', ${WP_DEBUG:-false});
     define('WP_DEBUG_LOG', ${WP_DEBUG_LOG:-false});
     define('WP_DEBUG_DISPLAY', ${WP_DEBUG_DISPLAY:-true});
-    if (\$_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https')
+    if (\$_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
        \$_SERVER['HTTPS']='on';
-
+    }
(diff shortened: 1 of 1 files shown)


dsifford:
Sorry for overlooking this @signaleleven!

I'm fairly certain this will be fine to merge in. I'm good with it if @karellm is.

karellm:
LGTM 👍

signaleleven:
Oh, that would be because I probably took the lines from 2 different files.
I am not used to maintaining a consistent style.
Should I commit a more consistent version? Which one would you prefer? I am more for the `{}` even for single line ifs...

karellm:
My preference is that a block should always be closed. So I definitely vote for `{}`.

dsifford:
@signaleleven Also, can you confirm that this doesn't throw warnings when not proxying just to be sure before we merge? I vaguely recall that I've had issues with using `$_SERVER` directly in some situations (can't recall exactly what those situations were though off the top of my head)

signaleleven:
I ran a site without a proxy in between and it looked ok. But since I use traefik for https and LE automation, I tested only pure http.

dsifford:
Support for this added in #130

PR #111, opened 2017-08-30
Instruction on how to build and use custom images

ieugen:
* also how to install extensions

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,57 @@ docker-compose up -d && docker-compose logs -f wordpress
 
 If you need a specific version, look at the [Changelog](CHANGELOG.md)
 
+#### Image customisation and php extensions
+
+When the default provided image is not enough or maybe missing some dependencies, please customise it. 
+The easiest way is to build a custom base image with a php extension starting from the base images provided.
+
+Let's add php `intl` to **php7.1** base image.
+We need to do the following:
+ * add the extension and dependencies to the image
+ * use the new image in our docker compose setup
+ 
+First we need to update **php7.1/Dockerfile** and update the build instructions to install the needed packages and dependencies for intl. Each extension has it's own dependencies. For the `intl` example we need to make the following changes: 
+
+php7.1/Dockerfile:
+````
+     && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+         less \
+         libpng12-dev \
+         libjpeg-dev \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,6 +59,57 @@ docker-compose up -d && docker-compose logs -f wordpress
...
+         
+````
+
+To use the new image we should also update **example/docker-compose.yml** to use build our image instead of using the published one.
+
+````
+ version: '3'
+ services:
+   wordpress:
+-    image: visiblevc/wordpress:0.15.2-php7.1
```
dsifford: This is not correct. The `image` key should only be used when pulling an image from a registry.

In this case, you're building your own image from scratch so it should instead look like this:

```yaml
version: '3'
services:
  wordpress:
    build: .  # Assuming the Dockerfile exists in the project root
# .... etc
```

README.md
```diff
@@ -59,6 +59,57 @@ docker-compose up -d && docker-compose logs -f wordpress
...
++    # image: visiblevc/wordpress:0.15.2-php7.1
++    build: ../php7.1
+     ports:
+````
+
+It may take multimple tryies until you figure out the correct software packages for each extension so you will have to rebuild your image again and again. To do that just run:
```
dsifford: Spelling

README.md
```diff
@@ -59,6 +59,57 @@ docker-compose up -d && docker-compose logs -f wordpress
...
+````
+   docker-compose build --no-cache
```
dsifford: Although this works, it's probably easier to use `docker-compose up --build` as it performs both the `build` and `up` steps.

karellm:
@dsifford I think that a dedicated wiki page could do the trick. We can link to it in the README. I agree on not crowding the README, this is quite specific.

ieugen:
Sure. As long as it is up.

dsifford:
@ieugen once you address the items that I brought up, I'd be glad to move the content of the PR to the repo's wiki

README.md
ieugen: That is a diff. I added the git diff in the documentation. 
The `image` line is marked as removed. That's how I thought it should be.

ieugen:
Should be ok. Thanks for the feedback.

dsifford:
Added contents of this PR to https://github.com/visiblevc/wordpress-starter/wiki/Community-Tips

Issue #110, opened 2017-08-24
Usability and readability - when errors happen with plugin install


carpnick:
## Overview

This issue is in regards to the output of `docker-compose up` when executing.  No OS version or docker version specific issues

The earlier output, earlier version of the container/run.sh, when installing plugins showed which plugin was having issues installing/activating, see sample:
```
wp_ebscoind.ebsco.com | ==> Checking themes
wp_ebscoind.ebsco.com |   -> (1/2) 'ebsco' listed as a local volume. SKIPPING... 
wp_ebscoind.ebsco.com |   -> (2/2) 'ebsco-child' listed as a local volume. SKIPPING... 
wp_ebscoind.ebsco.com | ==> Checking for orphaned themes
wp_ebscoind.ebsco.com |   -> 'twentyfifteen' no longer needed. Pruning ✓
wp_ebscoind.ebsco.com |   -> 'twentyseventeen' no longer needed. Pruning ✓
wp_ebscoind.ebsco.com |   -> 'twentysixteen' no longer needed. Pruning ✓
wp_ebscoind.ebsco.com | ==> Checking plugins
wp_ebscoind.ebsco.com |   -> (1/13) 'ajax-load-more' not found. Installing... ✓
wp_ebscoind.ebsco.com |   -> (2/13) 'akismet' found. SKIPPING... 
wp_ebscoind.ebsco.com |   -> (3/13) 'all-in-one-wp-security-and-firewall' not found. Installing... ✓
wp_ebscoind.ebsco.com |   -> (4/13) 'better-search-replace' not found. Installing... ✓
wp_ebscoind.ebsco.com |   -> (5/13) 'gravityforms' listed as a local volume. Activating... 
wp_ebscoind.ebsco.com |   -> (6/13) 'hyperdb' not found. Installing... ✓
wp_ebscoind.ebsco.com |   -> (7/13) 'instagram-feed' not found. Installing... ✓
wp_ebscoind.ebsco.com |   -> (8/13) 'mailgun' not found. Installing... ✓
wp_ebscoind.ebsco.com |   -> (9/13) 'share-this' not found. Installing... Error: No plugins installed.
wp_ebscoind.ebsco.com | ✘
wp_ebscoind.ebsco.com |   -> (10/13) 'types' not found. Installing... ✓
wp_ebscoind.ebsco.com |   -> (11/13) 'wordpress-importer' not found. Installing... ✓
[...]

PR #112, opened 2017-08-30
Add additional validation around WP commands

carpnick:
Provide validation around wp-cli commands and other output enhancements to let user/developer know when something failed, and enough verbose logging to see why it failed.

Could fix issue #110 

Sample outputs (all green):
image

Sample output (failed plugin):
image

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -83,6 +83,7 @@ h2() {
   echo -e "${ORANGE}${BOLD}==>${NC}${BOLD} $*${NC}"
 }
 
+
 _colorize() {
   local IN
   local success="${GREEN}${BOLD}Success:${NC}"
@@ -94,6 +95,15 @@ _colorize() {
     IN="${IN/Warning\:/$warning}"
     echo -e "$IN"
   done
+}
+
+_log_last_exit_colorize() {
+  if [ $? -eq 0 ]; then
+    echo "$1" |& _colorize
+  else
+    echo "$2" |& _colorize
+    exit 1
+  fi  
 }
 
 _get_volumes() {
@@ -115,7 +125,7 @@ _get_volumes() {
 }
 
 _wp() {
-  wp --allow-root "$@" |& _colorize
+  wp --allow-root "$@" 
 }
 
 # FIXME: Remove in next version
@@ -187,6 +197,7 @@ init() {
   if [[ ! -f /app/wp-settings.php ]]; then
     h2 "Downloading WordPress"
     _wp core download --version="$WP_VERSION"
+    _log_last_exit_colorize "Success: Wordpress downloaded" "Error: Wordpress download failed!"
   fi
 
   chown -R www-data /app/wp-content
@@ -199,15 +210,19 @@ check_database() {
   wp core is-installed --allow-root 2>/dev/null && return
 
   _wp db create
+  _log_last_exit_colorize "Success: db create" "Error: db create failed!"
 
   # No backups found
   if [[ "$( find /data -name "*.sql" 2>/dev/null | wc -l )" -eq 0 ]]; then
-    _wp core install
+    _wp core install 
+    _log_last_exit_colorize "Success: core install" "Error: core install failed!"
+
     return
   fi
 
   data_path=$( find /data -name "*.sql" -print -quit )
-  _wp db import "$data_path"
+  _wp db import "$data_path" 
+  _log_last_exit_colorize "Success: db import" "Error: db import failed!"
 
   if [[ -n "$URL_REPLACE" ]]; then
     wp search-replace --skip-columns=guid "$BEFORE_URL" "$AFTER_URL" --allow-root \
@@ -236,10 +251,15 @@ check_plugins() {
     to_remove+=( "$plugin" )
   done
 
-  [[ "${#to_install}" -gt 0 ]] && wp plugin install --allow-root "${to_install[@]}" | tail -n 1 |& _colorize
+  for key in "${to_install[@]}"; do
+    wp plugin install --allow-root "$key"
+    _log_last_exit_colorize "Success: $key plugin installed" "Error: $key plugin install failure!"
+  done  
+
   [[ "${#to_remove}" -gt 0 ]] && _wp plugin delete "${to_remove[@]}"
-  _wp plugin activate --all
-}
+  _wp plugin activate --all 
+  _log_last_exit_colorize "Success: plugin activate all" "Error: plugin activate all failed!"
+ }
 
 check_themes() {
   local key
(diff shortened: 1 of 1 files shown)


dsifford:
Thanks for the contribution @carpnick!

Sorry for the delay in getting to this. Gonna look through this now 😄

run.sh
```diff
@@ -255,15 +275,22 @@ check_themes() {
     done
   fi
 
-  [[ "${#to_install}" -gt 0 ]] && wp theme install --allow-root "${to_install[@]}" | tail -n 1 |& _colorize
+  for key in "${to_install[@]}"; do
```
dsifford: We used to do it this way (looping through each plugin individually), but we opted to lump them all together in the same command to speed up the build time.

Full confession: When we made this decision, we only assumed that wp-cli installs plugins concurrently. I've still never looked into it to find out.

Can you check on this and let me know? If wp-cli _does_ do the installs concurrently, can we try to go that route?

run.sh
```diff
@@ -255,15 +275,22 @@ check_themes() {
...
+  done 
 
   for theme in $(wp theme list --field=name --status=inactive --allow-root); do
     [[ ${theme_deps[$theme]} ]] && continue
     [[ ${theme_volumes[$theme]} ]] && continue
     to_remove+=( "$theme" )
   done
 
-  [[ "${#to_remove}" -gt 0 ]] && _wp theme delete "${to_remove[@]}"
+  for key in "${to_remove[@]}"; do
```
dsifford: Same as above

run.sh
carpnick: Plugin upgrades are **not** concurrent

If you follow the call stack: [wp plugin install](https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php#L707 ) calls into the [base class](https://github.com/wp-cli/extension-command/blob/master/src/WP_CLI/CommandWithUpgrade.php#L125) and does a [foreach](https://github.com/wp-cli/extension-command/blob/master/src/WP_CLI/CommandWithUpgrade.php#L128) slug

carpnick: I can still change if desired, just figured I would give you the first answer.

run.sh
carpnick: wp theme intall does the same thing.  It inherits from the same [base class](https://github.com/wp-cli/extension-command/blob/master/src/Theme_Command.php#L40) and calls the[ parent class install](https://github.com/wp-cli/extension-command/blob/master/src/Theme_Command.php#L508) the same way

carpnick:
If you dont like the logging, we can take a look at removing.  The problem as it currently stands, is when a command fails, the script continues executing.  If I fix just the exit code logic, it will at least fail for the user.

In the invalid plugin or theme use case, without the logging changes where we print wp-cli output, we cannot tell which plugin is failing to install, that was the reasoning.  

So really there are 2 pieces here: validating command success/error and printing wp-cli logging when errors happen to help the user understand what is failing.  I am open to other ways to get it there.

dsifford:
@carpnick Thanks for the claification.

Would simply setting `set -e` in script work for failing early? Do you know if a plugin or theme install failure when running the command with multiple plugins/themes results in a non-zero exit code?

dsifford:
@carpnick Yeah, good ole' bash! 😄 

Have you tried it to see? Didn't work? 

That would obviously be the easiest solution.

dsifford:
One issue that I can think of off the top of my head is circumstances where a database is imported and plugins are installed. 

If the site had, say, 2 of 3 plugins enabled and installed when the database was exported, wp-cli will report 1 of 3 plugins installed successfully, even though all 3 were installed.

carpnick:
It will **NOT** fail.  Example output with only the ```set -e``` change:

image

Run.sh inside container showing ```set -e```:

```
root@9e624f029056:/# cat run.sh |more -10
#!/bin/bash
# shellcheck disable=SC1091

#Error detection
set -e

# Environment
# ------------
ADMIN_EMAIL=${ADMIN_EMAIL:-"admin@${DB_NAME:-wordpress}.com"}
DB_HOST=${DB_HOST:-db}

```

dsifford:
Hmm.. 

I suppose this begs the question: Should the _entire_ build fail simply because a single plugin failed to install?

carpnick:
I am of the opinion... This is like programming, if any part of the `docker-compose up` fails, data-import, replacements, rewrites, plugin/theme install, rewrites... anything that is coded in that run.sh that could fail... should fail the build.

Otherwise, developer A and developer B could get two drastically different environments when running `docker-compose up`...

The only thing I have thought of, that might make this more palatable, is to add a FAILONERROR flag in the docker-compose.yml.  The default of the value is false, and a user needs to add it to their file to make it true.  If its true, we honor the [exit](https://github.com/visiblevc/wordpress-starter/pull/112/files#diff-1b0c2b516b83393edb7200ad5ff12181R105):
```
FAILONERROR=${FAILONERROR:-false}

...


_log_last_exit_colorize() {
  if [ $? -eq 0 ]; then
     echo "$1" |& _colorize
  else
    echo "$2" |& _colorize
    if [[ "$FAILONERROR" == 'true' ]]; then
      exit 1
    fi
 fi  
}
```

dsifford:
Good points. 

Thoughts @karellm?

My personal opinion: I'd rather just choose yes or no on the failure thing rather than add another environment variable flag.

karellm:
I also vote against the FLAG. 

My preference is to keep the build running and report the errors. Maybe the logger could store all errors and we could print a summary at this end to make it even more explicit?

carpnick:
If we need to create a logger and create a summary at the end, is this not adding more complexity that @dsifford just removed in a previous PR?

I am not sure why having it fail is not acceptable?  If there is any failure in bringing up the environment, why would we just not fail the container start?  It either brings up the development environment or it doesn't and prints an appropriate error message...

karellm:
As I said, I don't have a strong opinion on this. Both are fine with me. I differ to @dsifford for this.

carpnick:
@dsifford , thoughts?

carpnick:
As this is a sticking point for our developers, can we get a go/no-go on this?  Definitively.

I need to either fork permanently based on feedback from my developers or get this in.  We are about ready to roll this out to 100 or so WP sites we manage

dsifford:
@carpnick Sorry for the delay. Lots on my plate at the moment. I'll try to get back to this before the start of next week.

dsifford:
@carpnick Ok we can go ahead and merge this for now I suppose. 

The real solution I think will be to do a rewrite in python for better logging and control flow. Python is a strict dependency of the wordpress container anyway due to certbot so we might as well use it.

Adding that on the backburner for now.

carpnick:
Thank you.  Agreed, a more robust solution would probably be what we need.... Higher level language.

I will get the issue created for the python piece.  I have some experience with python and logging as we had to write an autopilot pattern recently that required extensive logging to figure out what was going on.

PR #121, opened 2017-12-03
add php7.2, remove deprecation warnings

### LINKED ISSUE #119: Add DB_USER environment variable support
## Overview

- What's your problem about?
There's no option to use another USER other than `root`
- What is your operating system?
MacOS
- What is your docker version? `docker version`
17.09.0-ce

## Project structure

```
/wp-content
  /plugins
    /my-plugin
  /themes
    /my-theme
```

dsifford:
cc: @karellm

**Questions:**
- Should we fully remove php5.6 from this repo? (My vote is yes -- historic dockerfiles will always exist on dockerhub).

**Notes:**
- The php7.2 image uses Debian Stretch, the others use Debian Jessie -- Updated the sources list accordingly.
- Another difference in the php7.2 image is the the PHP image uses the "slim" version of the distro (i.e., `debian:stretch-slim`). The others use the standard version (i.e. `debian:jessie`)

**TODO:**
- [x] Update README.md to include `DB_USER`
- [x] Version bump (minor semver)
 - [x] Build and push to dockerhub

**Future changes:** 
- I still think we should convert the run script into python, since we depend on it for certbot. Benefit is easier maintenance and also will allow us to run build steps concurrently. Planning on working on that as soon as time allows.
- In addition to above, I think it would be nice to come up with a better way to allow both configuration variables and configuration tasks to be extended without having to rebuild the base image. I have some ideas on how to do both of these things, but just writing this down in case you might have something better in mind.

Closes #119

Let me know what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ Thumbs.db
 *.swp
 .c9
 .sass-cache/
+.vscode
 
 # Wordpress
 wp-content/*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
-## 0.16.0 - latest
+## 0.17.0 - latest
+
+### Minor
+- Add PHP 7.2 (`latest`, `latest-php7.2`, `0.17.0-php7.2`)
+- Add `DB_USER` config option to environment variables
+
+## 0.16.0
 - **BREAKING CHANGE:** Builds will now exit if any plugin or theme installs fail.
 - Improve logging.
 
--- a/README.md
+++ b/README.md
@@ -1,39 +1,44 @@
 # Visible Wordpress Starter
 
-A Docker Wordpress development environment by the team at [Visible](https://visible.vc/) and some awesome [contributors](https://github.com/visiblevc/wordpress-starter/graphs/contributors). Our goal is to make Wordpress development slightly less frustrating.
+A Docker Wordpress development environment by the team at [Visible](https://visible.vc/) and some
+awesome [contributors](https://github.com/visiblevc/wordpress-starter/graphs/contributors). Our goal
+is to make Wordpress development slightly less frustrating.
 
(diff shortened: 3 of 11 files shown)


karellm:
LGTM 👍 

Regarding 5.6 we still use it at Visible but mostly out of lack of time/benefit to update. Eventually we will update the way the website is deployed but I don't know when that will be. What is the benefit of removing it beside lowering the "maintenance surface" ?

I also agree with both future changes! I don't say it often enough but this repo is really alive thanks to your great work! Thank you for that!

dsifford:
> What is the benefit of removing it beside lowering the "maintenance surface" ?

No huge benefit in removing. The fact that we have it just seems to me like it's an endorsement for using it, which I vehemently do not since it's now officially at "end of life" status and objectively slower than `7.x`.

> I also agree with both future changes! 

Great! Still tossing around the idea of how to implement to be honest. Not sure if it would be best to implement in python3 since we need it for certbot, OR if it would be better to implement it in something that can be compiled and injected as a binary to keep the size down (I'd have the most fun writing it in something like Rust, but there's several other languages in this space that I also wouldn't mind maintaining).

Still not certain certbot belongs in the wordpress image since it's really only useful on first build and then once every 3 months thereafter for renewals. Might move it to a separate image so that people who don't need it (or want it) don't get it in their wordpress image.

Just some ideas to mull over... Any thoughts?

>  I don't say it often enough but this repo is really alive thanks to your great work! Thank you for that!

Thanks for allowing me to collaborate :) -- The work I've done is mostly for selfish reasons (I use the image quite often!), but glad to hear you've also gotten use of it. Like I said way back in the beginning, I'm only here because your first image inspired me!

karellm:
> Just some ideas to mull over... Any thoughts?

I don't feel strongly here. I guess taking it outside will complicate the architecture without bringing big benefits.

PR #131, opened 2018-03-16
Add user on database test

luiscoms:


karellm:
Hi? Can you at least share your use case and why you need this for documentation purposes?

dsifford:
Not sure how the dyssynchrony happened here, but I checked in #130 and the user flag exists there. 

So we can probably close this.

luiscoms:
This is because #130 is not merged yet, you can merge this before #130

luiscoms:
The username on my database is not root, so, the mysqlping command fail

dsifford:
Sorry for the trouble. If you need a working implementation right this moment, you're free to fork the repo and build from the image directly.

If you can afford to wait just a bit longer, we'll be merging #130 soon. Building and pushing the images is non-trivial, so I'd prefer to avoid doing it twice in such a short amount of time.

Issue #132, opened 2018-04-22
Issue during wordpress installation


pramodskumbhar:
Hello,

I am following instructions and doing following:

```
git clone https://github.com/visiblevc/wordpress-starter.git
cd wordpress-starter/example
docker-compose up -d && docker-compose logs -f wordpress
```

and this is giving following error:

```
 (internal dummy connection)"
wordpress_1   | ======================================================================
wordpress_1   |                     Begin WordPress Installation
wordpress_1   | ======================================================================
wordpress_1   | ==> Waiting for MySQL to initialize...
wordpress_1   | ==> Configuring WordPress
wordpress_1   |          ERROR 1045 (28000): Access denied for user
wordpress_1   |          'root'@'172.18.0.4' (using password: YES)
wordpress_1   | ==> Checking database
wordpress_1   |   Error: 'wp-config.php' not found.
wordpress_1   |          Either create one manually or use `wp config create`.
wordpress_1   |   Error: 'wp-config.php' not found.
wordpress_1   |          Either create one manually or use `wp config create`.
wordpress_1   | ==> Checking themes
wordpress_1   | Error: 'wp-config.php' not found.
wordpress_1   | Either create one manually or use `wp config create`.
wordpress_1   | Error: 'wp-config.php' not found.
wordpress_1   | Either create one manually or use `wp config create`.
wordpress_1   |   Error: 'wp-config.php' not found.
wordpress_1   |          Either create one manually or use `wp config create`.
wordpress_1   | ==> Checking plugins
wordpress_1   | Error: 'wp-config.php' not found.
wordpress_1   | Either create one manually or use `wp config create`.
wordpress_1   | Error: 'wp-config.php' not found.
wordpress_1   | Either create one manually or use `wp config create`.
wordpress_1   |   Error: 'wp-config.php' not found.
wordpress_1   |          Either create one manually or use `wp config create`.
wordpress_1   | Error: 'wp-config.php' not found.
wordpress_1   | Either create one manually or use `wp config create`.
wordpress_1   | ==> Finalizing
wordpress_1   |   Error: 'wp-config.php' not found.
wordpress_1   |          Either create one manually or use `wp config create`.
wordpress_1   | ======================================================================
wordpress_1   |                   WordPress Configuration Complete!
wordpress_1   | ======================================================================
[...]

PR #136, opened 2018-05-03
fix permissions issues once and for all with bindfs

dsifford:
### Intro

The purpose of this PR is to solve the longstanding permissions issues that have existed when working with files/folders needing to be shared between the admin user (`admin`) and the server user (`www-data`).

Currently, and in all iterations in the past, whenever a user opened a tty session in the wordpress container and then manipulated any files manually, or with wp-cli, the permissions of those files would modify from being owned by `www-data` to being owned by `admin`. This becomes an issue if those files then needed subsequent modification by the server.

A common example of this happening is when a user installs a new plugin with `wp plugin instal XXX`, the server could then not update or remove the plugin itself from the wordpress dashboard (due to improper permissions).

### What this PR does

To combat these permissions issues once and for all, I opted to implement a [bindfs](http://bindfs.org) mount to the webroot (`/var/www/html`) from the `/app` directory.

Bindfs is a FUSE based filesystem that allows a filesystem on a system to be "re-mounted" virtually in another location on the same system. This has useful implications for us because by using bindfs, we're able to "re-mount" the `/app` directory to the webroot and **force** all the files in the webroot directory to be owned by `www-data`.

What's nice about this also is that if the server creates any files in the webroot, those files are synced backwards to the `/app` directory and permissions are then changed back to "owner:admin group:admin".

### TL;DR

This PR solves the permissions issues permanently.

### BREAKING CHANGE NOTE:

The WordPress service must now be configured in one of the following ways:

(preferred)
```yml
services:
    wordpress:
        cap_add:
            - SYS_ADMIN
        devices:
            - /dev/fuse
```

**OR**

(try to avoid if possible)
```yml
services:
    wordpress:
        privileged: true
```

cc: @karellm

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,6 +1,31 @@
 #!/bin/bash
 # shellcheck disable=SC1091
 
+if ! sudo mount -a 2>/dev/null; then
+    printf '\e[1;31mERROR:\e[0m %s' \
+    'Container running with improper privileges.
+
+    Be sure your service is confiured with the following options:
+    ___
+    services:
+    wordpress:
+        cap_add:
+        - SYS_ADMIN
+        devices:
+        - /dev/fuse
+    ___
+
+    OR (use first option if possible)
+    ___
+    services:
+    wordpress:
+        privileged: true
+    ___
+
+    ' | sed 's/^    //'
+    exit 1
+fi
+
 # Environment
 # ------------
 declare -x TERM="${TERM:-xterm}"
@@ -94,10 +119,6 @@ main() {
         wp --color rewrite structure \
             "${PERMALINKS:-/%year%/%monthnum%/%postname%/}" |& logger
     fi
-
-    chgrp -R www-data /app
-    chmod -R g+rw /app
-    find /app -type d -exec chmod g+s {} +
 
     h1 'WordPress Configuration Complete!'
 
--- a/run.sh
+++ b/run.sh
@@ -5,7 +5,7 @@ if ! sudo mount -a 2>/dev/null; then
     printf '\e[1;31mERROR:\e[0m %s' \
     'Container running with improper privileges.
 
-    Be sure your service is confiured with the following options:
+    Be sure your service is configured with the following options:
     ___
     services:
     wordpress:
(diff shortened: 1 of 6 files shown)


dsifford:
Of note: I've been using this for the last 3 days without issue.

dsifford:
Of note, of note: Still using this. It's been a dream since day 1. No probs at all and all headaches gone.

Let me know when we're okay to merge. @karellm

run.sh
```diff
@@ -1,6 +1,31 @@
 #!/bin/bash
 # shellcheck disable=SC1091
 
+if ! sudo mount -a 2>/dev/null; then
+    printf '\e[1;31mERROR:\e[0m %s' \
+    'Container running with improper privileges.
+
+    Be sure your service is confiured with the following options:
```
karellm: configured is mispelled

dsifford:
Let me know what you'd prefer to do Re: showing the configuration settings. If you'd prefer to just show the 1, that's fine.

Once I hear back I'll go ahead and merge / push out. :+1:

karellm:
@dsifford I trust your judgement on this one. LGTM 👍

dsifford:
You have the `cap_add` and `devices` options enabled and it's still giving you those issues?

dsifford:
Can you paste over your compose file?

Vantablack:
I'm using the latest one which is https://github.com/visiblevc/wordpress-starter/blob/d909f9e1b488e6e5648ba084330aac6f1b62e83b/example/docker-compose.yml

Vantablack:
Hey @dsifford it seems like adding `--security-opt apparmor:unconfined` worked (on Ubuntu)

```yaml
...
services:
  wordpress:
    # required for mounting bindfs
    cap_add:
      - SYS_ADMIN
    devices:
      - /dev/fuse
    security_opt:
      - apparmor:unconfined
...
```

The solution was from https://github.com/moby/moby/issues/16429#issuecomment-217126586 and this issue was also mentioned in https://github.com/visiblevc/wordpress-starter/issues/132#issuecomment-390498838

karellm:
@dsifford or @Vantablack could you make a PR documenting this in the README?

PR #139, opened 2018-05-25
add redis support

dsifford:
This PR simply adds support for Redis object caching. Nothing particularly interesting or fancy.

cc: @karellm 

**Edit:** The reason for the tweaks to the zip configuration is because there was a warning message that showed up during build unrelated to the redis changes that stated libzip was being removed as a dependency of PHP soon and needed to be installed manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
-## 0.19.0 - latest
+## 0.20.0 - latest
+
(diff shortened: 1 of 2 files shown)


dsifford:
@karellm Re: adding note about charset support in changelog -- will do!

karellm:
My bad, I didn't read it carefully. All good!

PR #145, opened 2018-07-26
Update dump command example in README.md

PabloDinella:
Also would like to say that this project is awesome, thank you for making it available.


dsifford:
Thanks!

PR #147, opened 2018-08-23
perform db update on import + improve docs

dsifford:
This PR adds an extremely minor change that mitigates a small annoyance.

Typically, if a user imports an existing database into a `nightly` wordpress site, when logging it they'll be hit with a "you must update your database to continue" message. This PR adds a check that automatically does this on build.

This PR also fixes minor documentation issues.

Once approved, I'll build targeting the latest builds of each PHP minor version that we're tracking (we're behind a few).

cc: @karellm

### DIFF
--- a/example/docker-compose.yml
+++ b/example/docker-compose.yml
@@ -1,38 +1,41 @@
-version: '3'
+version: "3"
 services:
-  wordpress:
-    image: visiblevc/wordpress
+    wordpress:
+        image: visiblevc/wordpress
 
-    # required for mounting bindfs
-    cap_add:
-      - SYS_ADMIN
-    devices:
-      - /dev/fuse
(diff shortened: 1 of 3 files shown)


run.sh
```diff
@@ -191,6 +194,8 @@ check_database() {
             "$(wp option get siteurl)" \
             "$URL_REPLACE" |& logger
     fi
+
+    wp --color core update-db |& logger
```
dsifford: This is the only line that actually changes the behavior of the image.

example/docker-compose.yml
```diff
@@ -1,38 +1,41 @@
-version: '3'
+version: "3"
```
dsifford: formatted this file with prettier. that's the reason for the whitespace changes.

PR #149, opened 2018-11-04
add SSL note about setting SERVER_NAME before obtaining certificate

PabloDinella:
This can avoid frustration when not being able to obtain SSL certificates


dsifford:
Thanks

PR #150, opened 2018-12-06
add /docker-entrypoint-initwp.d/ init scripts directory

dsifford:
This PR adds a small feature that simply follows the format of how MySQL and MariaDB have for allowing users to run arbitrary init scripts after build completion. Let me know if you'd like me to explain further and I'd be glad to do so.

I added an example into the `/examples` directory of this functionality in action.

Also, this bumps the default theme to `twentynineteen`.

Ping: @karellm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Minor
 
+- Add `php7.3`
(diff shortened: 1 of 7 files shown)


dsifford:
Also I think PHP 7.3 is releasing today? If so, I'll add that too....

dsifford:
Ping @karellm

karellm:
LGTM 👍 Thanks for your patience!

PR #154, opened 2019-05-11
add missing site health extensions

dsifford:
WordPress 5.2 now comes with a "Site Health" feature. Currently, this feature prints out a warning due to the fact that the image didn't have `bcmath` or `imagick` PHP extensions installed.

This PR installs and enables those extensions, which fixes those warnings.

Let me know if you have any questions.

Ping: @karellm

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,22 @@
-ARG PHP_VERSION=7.2
+ARG PHP_VERSION=7.3
 FROM php:${PHP_VERSION}-apache
 ARG VERSION=latest
 LABEL maintainer="Derek P Sifford <dereksifford@gmail.com>" \
       version="${VERSION}-php${PHP_VERSION}"
 
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -53,8 +61,7 @@ RUN echo "deb http://ftp.debian.org/debian $(sed -n 's/^VERSION=.*(\(.*\)).*/\1/
         -o /usr/local/bin/wp https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar \
         -o /etc/bash_completion.d/wp-cli https://raw.githubusercontent.com/wp-cli/wp-cli/master/utils/wp-completion.bash \
     && a2enconf fqdn \
-    && a2enmod rewrite expires \
-    && service apache2 restart
```
karellm: Isn't it needed anymore?

dsifford: Wasn't ever needed in the first place. We start apache at the end of `run.sh`.

karellm: Makes total sense. LGTM!

PR #165, opened 2019-11-19
greatly simplify runtime script

dsifford:
This PR seeks to greatly simplify the process of bootstrapping a wordpress installation from how we've done it in the past.

Previously, there was no way to do a clean install of wordpress without default plugins and themes, so we did a really complex method of checking for dependencies and then went through and manually deletes plugins/themes that didn't check out. There are several problems/quirks with doing it this way and I think we've talked about them in the past and just delt with them.

Now wp-cli allows you to download wordpress without the `wp-content` directory. So there really is no need to do this convoluted volume check anymore.

So that's the biggest change I made. I did another quick once over and made sure everything else checked out still and adjusted misc stuff during that little process (I'll look and comment on notable stuff inline in this PR if there are any).

Ping @karellm

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -34,20 +34,19 @@ fi
 declare -x TERM="${TERM:-xterm}"
 declare PLUGINS="${PLUGINS//,/}"
 declare THEMES="${THEMES//,/}"
+
+# Runtime
+# ------------
+declare default_theme=twentytwenty
 declare -A plugin_deps
 declare -A theme_deps
-
-# FIXME: Deprecation of old version of $URL_REPLACE
-if [[ $URL_REPLACE =~ , ]]; then
-    URL_REPLACE=${URL_REPLACE%%,*}
-    printf '\e[1;33;7mDEPRECATED\e[0m %s\n' "URL_REPLACE must only contain AFTER_URL. BEFORE_URL,AFTER_URL form has been deprecated"
-fi
 
 # Configuration
 # -------------
 mkdir -p ~/.wp-cli
 echo -e "
 path: /app
+color: true
 apache_modules:
     - mod_rewrite
 
@@ -82,13 +81,6 @@ rewrite structure:
 
 " > ~/.wp-cli/config.yml
 
-# Apache config adustments
-sudo sed -i \
-    -e "/^[[:blank:]]*.ServerName www.example.com/{c\\" \
-    -e "\\tServerName ${SERVER_NAME:-localhost} \\" \
-    -e "\\tServerAlias www.${SERVER_NAME:-localhost}" \
-    -e '}' /etc/apache2/sites-available/000-default.conf
-
 main() {
     h1 'Begin WordPress Installation'
     init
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -132,7 +124,7 @@ main() {
         done
     fi
 
-    h1 'WordPress Configuration Complete!'
+    h1 'WordPress Installation Complete!'
```
dsifford: I've had two people I work with comment about the language here. So I switched it to match the initial `h1` language.

run.sh
```diff
@@ -82,13 +81,6 @@ rewrite structure:
 
 " > ~/.wp-cli/config.yml
 
-# Apache config adustments
-sudo sed -i \
-    -e "/^[[:blank:]]*.ServerName www.example.com/{c\\" \
-    -e "\\tServerName ${SERVER_NAME:-localhost} \\" \
-    -e "\\tServerAlias www.${SERVER_NAME:-localhost}" \
-    -e '}' /etc/apache2/sites-available/000-default.conf
-
```
dsifford: Moved into the init function

run.sh
```diff
@@ -170,160 +160,69 @@ init() {
     done
 
     # If no theme dependencies or volumes exist, fall back to default
-    if [[ ${#theme_deps[@]} == 0 && $(check_volumes -t) == "" ]]; then
-        theme_deps[twentytwenty]=twentytwenty
+    if [[ ${#theme_deps[@]} == 0 && -d /app/wp-content/themes && $(find /app/wp-content/themes/* -maxdepth 0 -t d | wc -l) == 0 ]]; then
+        theme_deps["$default_theme"]="$default_theme"
```
dsifford: Makes it easier to update this in the future since it's defined up at the top.

run.sh
```diff
@@ -170,160 +160,69 @@ init() {
...
-    declare data_path
-    data_path=$(find /data -name '*.sql' -print -quit 2> /dev/null)
-    if [[ ! "$data_path" ]]; then
-        wp --color core install |& logger
-        return
-    fi
+    wp db create |& logger
 
-    wp --color db import "$data_path" |& logger
+    for file in /data/*.sql; do
```
dsifford: Allows users to import more than one database file (in the event they download chunks of a single installation) -- I'm running into the need for this so I figured I'd add it in.

dsifford:
It should probably be noted that we've had the ability to do it in this simplified way for what appears to be [2 entire years](https://github.com/visiblevc/wordpress-starter/blame/99d60c5b93f1a6cddd55f0c7f08d44168fd07146/run.sh#L69) but for whatever reason I must have enjoyed making more work for myself.

run.sh
```diff
@@ -187,22 +187,26 @@ check_database() {
 
     wp db create |& logger
 
+    wp core install |& logger
```
dsifford: Install wordpress no matter what, even if importing databases (just to ensure we have all the tables if doing a partial import).

dsifford:
I just built and tested the images and it works as expected. So I'm gonna go ahead and merge this and push them out because I'm low on time.

If you're not fond of the last min changes, I'll roll back and push another version. But I can't think of any reason why you wouldn't be.

PR #167, opened 2019-12-07
fix small bug in run.sh + drop support for php < 7.2 + update CHANGELOG

dsifford:
This PR fixes a small bug in run.sh resulting from a typo in a `find` flag.

Additionally, this PR drops support for PHP < 7.2 to [align with php](https://www.php.net/supported-versions.php)

Finally, the build script was updated to remove useless cruft and the CHANGELOG was brought up to date.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,22 +3,13 @@ set -e
 
 : "${npm_package_version?Script must be ran using npm}"
 
-# Ascending order is important here
-declare -a php_versions=(
-    7.2
-    7.3
-    7.4
-)
-
-# NOTE: Not building this stack of images concurrently due to a known issue
-# with docker concurrent builds. https://github.com/moby/moby/issues/9656
-for php_version in "${php_versions[@]}"; do
+for dir in ./images/*; do
     docker build \
-        --build-arg PHP_VERSION="$php_version" \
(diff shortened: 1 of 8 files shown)


build.sh
```diff
@@ -21,7 +19,7 @@ for php_version in "${php_versions[@]}"; do
         -t "visiblevc/wordpress:latest" \
         -t "visiblevc/wordpress:latest-php${php_version}" \
         -t "visiblevc/wordpress:$npm_package_version-php${php_version}" \
-        "$dockerfile_dir"
+        .
```
dsifford: Not needed since running using npm will ensure that the `$PWD` is the root of the repo.

dsifford:
Okay, so this got a bit more complicated... 

It appears the configuration for php7.4 extensions has diverged from versions below it, so we're no longer able to share dockerfiles anymore.

See [here](https://github.com/docker-library/php/issues/920#issuecomment-562864296) for an example of this. There are other such cases as well.

So have a look at this and let me know what you think @karellm

dsifford:
Sorry for the delay. Sounds good.

We've been using this at my workplace so I'm able to find and correct issues relatively quickly. I'll be sure to do so if we notice any probs with this.

Issue #151, opened 2018-12-17
mysql:5.7 hangs on waiting for MySQL to initialize


martintburgess:
## Overview
Thanks for a great repo. 
I am trying to run /example/docker-compose.yml using mysql:5.7 instead of mariadb:10
This is the only change to the docker-compose.yml.
Installation doesn't get passed "Waiting for MySQL to initialize...".

Issue does not occur with mariadb:10

mac os x 10.14.2
docker version 18.09.0

## `docker-compose.yml`

```yml
version: "3"
services:
    wordpress:
        image: visiblevc/wordpress

        # required for mounting bindfs
        cap_add:
            - SYS_ADMIN
        devices:
            - /dev/fuse
        # required on certain cloud hosts
        security_opt:
            - apparmor:unconfined

        ports:
            - 8080:80
            - 443:443
        volumes:
            - ./data:/data
            - ./scripts:/docker-entrypoint-initwp.d
        environment:
            DB_NAME: wordpress
            DB_PASS: root
            PLUGINS: >-
                academic-bloggers-toolkit
                co-authors-plus
                [WP-API]https://github.com/WP-API/WP-API/archive/master.zip

    db:
        image: mysql:5.7 # mariadb:10 
        volumes:
            - data:/var/lib/mysql
        environment:
            MYSQL_ROOT_PASSWORD: root

    phpmyadmin:
        image: phpmyadmin/phpmyadmin
        ports:
            - 22222:80

volumes:
    data:
```

## Project structure
same as /example

## `docker-compose up` output

```
wordpress_1   | ======================================================================
wordpress_1   |                     Begin WordPress Installation                    
wordpress_1   | ======================================================================
wordpress_1   | ==> Downloading WordPress
wordpress_1   |          Downloading WordPress 5.0.1 (en_US)...
wordpress_1   |          md5 hash verified: ae286aad683f26c19ce3e4cff91b5661
wordpress_1   | Success: WordPress downloaded.
wordpress_1   | ==> Waiting for MySQL to initialize...
```

dsifford:
@martintburgess Can you explain further what the issue is you're having. The output I'm seeing looks correct.

martintburgess:
@dsifford the installation doesn't get past this point. Waited considerable time (15 minutes or more) and was stuck there.

dsifford:
Ah... ok got it. Not sure why, but I will confess that I generally always use MariaDB... I'll check into this tomorrow.

[...]

PR #171, opened 2020-01-26
fix: add info about db port

jiverson:


dsifford:
Hey @jiverson thanks for the PR.

I'm wondering how this change would be helpful? Since we're not exposing port `3306` outside of the docker network, exposing it as 3307 shouldn't have any effect other thank just allowing you to access it from the host machine.

jiverson:
Yeah it is just there for ones who had experienced this issue: https://github.com/visiblevc/wordpress-starter/issues/151#issuecomment-578519554 

If you would like to instead add it to the README or something up to you. But I was experiencing another issue which I am addressing in another PR I am working on. 

At the time I wasn't sure what exactly was going on.

dsifford:
So looking at your comments in #172 ....

> I thought maybe it was my db so I switched to mysql which then would not initialize (obviously I had local mysql already using that port #171).

Even if you have mysql operating on your host on port 3306, mysql in the container will work just fine because it doesn't map to 3306 on the host, just in the container and its associated docker network. So changing mapping to 3307 on the host shouldn't have any affect on the init process.

jiverson:
Yeah I can close this not a big deal honestly.

PR #172, opened 2020-01-26
fix: themes folder

jiverson:
So I am no expert in wordpress nor a bash expert.

Two things to note was that I did not add a theme with `THEMES` in the `[theme.slug]` parameter and I also did not add a custom theme with plugins. 

My `docker-compose.yml` looked exactly like the example.

What I found out is though after install, no default theme was applied, which from what I could tell was suppose to fallback to `twentytwenty`. But even after logging into wp-admin I could not install any themes. The error was something along the lines of the folder could not be created.

I thought maybe it was my db so I switched to mysql which then would not initialize (obviously I had local mysql already using that port https://github.com/visiblevc/wordpress-starter/pull/171).

As far as I can tell `wp cli` does not create the `themes` folder in `wp-content/` on install, possibly because of `skip-content`? Because of that the if statement would never install the default theme.

It just seems to me that the `themes` folder needs to be created with possibly the right permissions? I am not 100% sure but this does the fix the issue, feel free to reject if you find a better way.

The fix was to add another check to see if the `app/wp-content/themes` directory does not exist at fallback to the default theme.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -39,6 +39,8 @@ services:
 
     db:
         image: mariadb:10 # or mysql:5.7
+        ports:
+            - 3307:3306 # change ip if required
         volumes:
             - data:/var/lib/mysql
         environment:
--- a/example/README.md
+++ b/example/README.md
@@ -39,8 +39,6 @@ services:
 
     db:
         image: mariadb:10 # or mysql:5.7
-        ports:
-            - 3307:3306 # change ip if required
         volumes:
             - data:/var/lib/mysql
         environment:
--- a/example/docker-compose.yml
+++ b/example/docker-compose.yml
@@ -28,6 +28,8 @@ services:
 
     db:
         image: mariadb:10 # or mysql:5.7
+        ports:
+            - 3307:3306 # change ip if required
         volumes:
             - data:/var/lib/mysql
         environment:
--- a/example/docker-compose.yml
+++ b/example/docker-compose.yml
@@ -28,8 +28,6 @@ services:
 
     db:
         image: mariadb:10 # or mysql:5.7
-        ports:
-            - 3307:3306 # change ip if required
         volumes:
             - data:/var/lib/mysql
         environment:
--- a/run.sh
+++ b/run.sh
@@ -160,7 +160,8 @@ init() {
     done
 
     # If no theme dependencies or volumes exist, fall back to default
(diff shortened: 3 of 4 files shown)


example/README.md
```diff
@@ -39,6 +39,8 @@ services:
 
     db:
         image: mariadb:10 # or mysql:5.7
+        ports:
+            - 3307:3306 # change ip if required
```
dsifford: Remove this since it's out of scope for this PR.

example/docker-compose.yml
```diff
@@ -28,6 +28,8 @@ services:
 
     db:
         image: mariadb:10 # or mysql:5.7
+        ports:
+            - 3307:3306 # change ip if required
```
dsifford: Remove this since it's out of scope for this PR.

run.sh
```diff
@@ -160,7 +160,8 @@ init() {
     done
 
     # If no theme dependencies or volumes exist, fall back to default
-    if [[ ${#theme_deps[@]} == 0 && -d /app/wp-content/themes && $(find /app/wp-content/themes/* -maxdepth 0 -type d | wc -l) == 0 ]]; then
+    if [[ (${#theme_deps[@]} == 0 && ! -d /app/wp-content/themes) ||
+            (-d /app/wp-content/themes && $(find /app/wp-content/themes/* -maxdepth 0 -type d | wc -l) == 0)]]; then
```
dsifford:
```suggestion
```

run.sh
```diff
@@ -160,7 +160,8 @@ init() {
     done
 
     # If no theme dependencies or volumes exist, fall back to default
-    if [[ ${#theme_deps[@]} == 0 && -d /app/wp-content/themes && $(find /app/wp-content/themes/* -maxdepth 0 -type d | wc -l) == 0 ]]; then
+    if [[ (${#theme_deps[@]} == 0 && ! -d /app/wp-content/themes) ||
```
dsifford: The condition below should be all that we need to fix the issue here.

```suggestion
    if [[ (${#theme_deps[@]} == 0 && ! -d /app/wp-content/themes) ]]; then
```

example/docker-compose.yml
jiverson: Done

jiverson:
Sounds good to me thanks for the quick feedback.

PR #182, opened 2021-05-26
Remove duplicate phpMyAdmin service from example docs

AlecRust:
The example `docker-compose.yml` in `/example/README.md` already includes phpMyAdmin. But then a separate `docker-compose.yml` example to enable phpMyAdmin is listed, which is the same except `MYSQL_ROOT_PASSWORD` is added, which is not documented in main README under "Service Environment Variables" for `phpmyadmin` and not included in [the actual](https://github.com/visiblevc/wordpress-starter/blob/master/example/docker-compose.yml) example `docker-compose.yml`.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -44,6 +44,7 @@ services:
         environment:
             MYSQL_ROOT_PASSWORD: root
 
+    # if you want phpMyAdmin
     phpmyadmin:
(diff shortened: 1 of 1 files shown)


dsifford (approved):

dsifford:
Thanks!

